PR #5, opened 2024-11-21
ci: add linting job

sadlerap:
This adds a job in github actions to lint go code using golangci-lint. It adds a configuration to enable more checks than the default, so some linter checks we would want (such as gosec) won't need to be specified with a separate job.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,37 @@
+name: Server CI
+
+on:  # yamllint disable-line rule:truthy
+  pull_request: {}
+  push:
+    branches: [main]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  lint-go:
+    name: Lint code
+    runs-on: ubuntu-24.04
+
+    steps:
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: "1.22"
+
+      - name: Checkout Git Repository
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,37 @@
+name: Server CI
+
+on:  # yamllint disable-line rule:truthy
+  pull_request: {}
```
filariow, line 4:
```suggestion
  pull_request_target: {}
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+      - name: Checkout Git Repository
+        uses: actions/checkout@v4
+
+      - name: Determine golang-ci version
+        id: golangci_version
+        run: |
+          echo "version=$(go mod edit -json hack/tools/golang-ci/go.mod | \
+            jq '.Require | map(select(.Path == "github.com/golangci/golangci-lint"))[].Version')" \
+            >> $GITHUB_OUTPUT
```
filariow, line 31: interesting, we could do the same thing for the Go Version

```
go mod edit -json  | jq '.Go'
```

filariow (changes requested):
mostly LGTM, just some suggetions on the pipeline. 

PR needs a rebase

.github/workflows/ci.yaml
sadlerap: Why do we need to target `pull_request_target`?  I don't think we would stand to gain from the extra privileges that `pull_request_target` would give us in this action.

filariow (approved):
lgtm

PR #6, opened 2024-11-21
Update k8s.io/apiserver from v0.31.0 to v0.31.2

sadlerap:
All the other dependencies from the `k8s.io` group are also at `v0.31.2`, so this should be as well.


sadlerap:
Open question: do we want to use dependabot?  We could also use Konflux to keep our dependencies up-to-date and dogfood our own product.

filariow (approved):
As this PR is ready and good to merge, I'd say let's use dependabot short-term, Konflux long-term

filariow (approved):
lgtm

PR #29, opened 2025-01-16
Allow TLS enablement

sadlerap:
This PR allows TLS to be enabled via the addition of three command-line arguments:
```
  -cert-path string
        Path to TLS certificate store.
  -enable-tls
        Toggle tls enablement (default true)
  -key-path string
        Path to TLS private key.
```

TLS is enabled by default, but disable it during testing to allow for easier debugging.

### DIFF
--- a/http_server.go
+++ b/http_server.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"crypto/tls"
 	"log/slog"
 	"net/http"
 	"os"
@@ -16,6 +17,8 @@ const (
 
 type NamespaceListerServer struct {
 	*http.Server
+	useTLS  bool
+	tlsOpts []func(*tls.Config)
 }
 
 func addInjectLoggerMiddleware(l *slog.Logger, next http.Handler) http.HandlerFunc {
@@ -77,6 +80,16 @@ func NewServer(l *slog.Logger, ar authenticator.Request, lister NamespaceLister)
 	}
 }
 
+func (s *NamespaceListerServer) WithTLS(enableTLS bool) *NamespaceListerServer {
+	s.useTLS = enableTLS
+	return s
+}
+
+func (s *NamespaceListerServer) WithTLSOpts(tlsOpts ...func(*tls.Config)) *NamespaceListerServer {
+	s.tlsOpts = tlsOpts
+	return s
+}
+
 func (s *NamespaceListerServer) Start(ctx context.Context) error {
 	// HTTP Server graceful shutdown
 	go func() {
@@ -91,6 +104,17 @@ func (s *NamespaceListerServer) Start(ctx context.Context) error {
 			os.Exit(1)
 		}
 	}()
+
+	// setup and serve over TLS if configured
+	if s.useTLS {
+		s.TLSConfig = &tls.Config{
+			MinVersion: tls.VersionTLS12,
+		}
+		for _, fun := range s.tlsOpts {
+			fun(s.TLSConfig)
+		}
+		return s.ListenAndServeTLS("", "")
+	}
 
 	// start server
 	return s.ListenAndServe()
--- a/main.go
+++ b/main.go
@@ -2,6 +2,9 @@ package main
 
 import (
 	"context"
+	"crypto/tls"
+	"flag"
+	"fmt"
 	"log/slog"
 	"os"
 	"os/signal"
@@ -21,8 +24,32 @@ func main() {
 	}
 }
 
+func loadTLSCert(l *slog.Logger, certPath, keyPath string) func(*tls.Config) {
+	getCertificate := func(_ *tls.ClientHelloInfo) (*tls.Certificate, error) {
+		cert, err := tls.LoadX509KeyPair(certPath, keyPath)
+		if err != nil {
+			l.Error("Unable to load TLS certificates", "error", err)
+			return nil, fmt.Errorf("Unable to load TLS certificates: %w", err)
+		}
+
+		return &cert, err
+	}
+
+	return func(config *tls.Config) {
+		config.GetCertificate = getCertificate
+	}
+}
+
 func run(l *slog.Logger) error {
 	log.SetLogger(logr.FromSlogHandler(l.Handler()))
+
+	var enableTLS bool
+	var tlsCertificatePath string
+	var tlsCertificateKeyPath string
+	flag.BoolVar(&enableTLS, "enable-tls", true, "Toggle TLS enablement.")
+	flag.StringVar(&tlsCertificatePath, "cert-path", "", "Path to TLS certificate store.")
+	flag.StringVar(&tlsCertificateKeyPath, "key-path", "", "Path to TLS private key.")
+	flag.Parse()
 
 	// get config
 	cfg := ctrl.GetConfigOrDie()
@@ -56,6 +83,10 @@ func run(l *slog.Logger) error {
 	// build http server
 	l.Info("building server")
 	s := NewServer(l, ar, nsl)
+
+	// configure TLS
+	s.WithTLS(enableTLS).
+		WithTLSOpts(loadTLSCert(l, tlsCertificatePath, tlsCertificateKeyPath))
 
 	// start the server
 	return s.Start(ctx)
--- a/config/deployment.yaml
+++ b/config/deployment.yaml
@@ -19,7 +19,9 @@ spec:
       #   runAsNonRoot: true
       serviceAccountName: namespace-lister
       containers:
-      - image: namespace-lister:latest
+      - args:
+        - -enable-tls=false
+        image: namespace-lister:latest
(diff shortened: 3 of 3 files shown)


http_server.go
```diff
@@ -16,6 +17,8 @@ const (
 
 type NamespaceListerServer struct {
 	*http.Server
+	useTls  bool
```
filariow, line 20:
```suggestion
	useTLS  bool
```

IIRC, by convention this should be uppercase

main.go
```diff
@@ -21,9 +24,32 @@ func main() {
...
+	}
+}
+
 func run(l *slog.Logger) error {
 	log.SetLogger(logr.FromSlogHandler(l.Handler()))
 
+	var enableTLS bool
+	var tlsCertificatePath string
+	var tlsCertificateKeyPath string
+	flag.BoolVar(&enableTLS, "enable-tls", true, "Toggle tls enablement")
```
filariow, line 48:
```suggestion
	flag.BoolVar(&enableTLS, "enable-tls", true, "Toggle TLS enablement")
```

main.go
```diff
@@ -21,9 +24,32 @@ func main() {
...
+	flag.StringVar(&tlsCertificatePath, "cert-path", "", "Path to TLS certificate store.")
+	flag.StringVar(&tlsCertificateKeyPath, "key-path", "", "Path to TLS private key.")
```
filariow, line 50:
```suggestion
	flag.StringVar(&tlsCertificatePath, "cert-path", "", "Path to TLS certificate store")
	flag.StringVar(&tlsCertificateKeyPath, "key-path", "", "Path to TLS private key")
```

main.go
```diff
@@ -57,6 +83,15 @@ func run(l *slog.Logger) error {
 	l.Info("building server")
 	s := NewServer(l, ar, nsl)
 
+	// configure TLS
+	if enableTLS {
+		s.WithTLS(true)
+		tlsOpts := []func(*tls.Config){
+			loadTLSCert(l, tlsCertificatePath, tlsCertificateKeyPath),
+		}
+		s.WithTLSOpts(tlsOpts)
```
filariow, line 92: This seems redundant to me


```suggestion
		tlsOpts := []func(*tls.Config){
			loadTLSCert(l, tlsCertificatePath, tlsCertificateKeyPath),
		}
		s.WithTLSOpts(tlsOpts)
```

http_server.go
```diff
@@ -77,6 +80,16 @@ func NewServer(l *slog.Logger, ar authenticator.Request, lister NamespaceLister)
 	}
 }
 
+func (s *NamespaceListerServer) WithTLS(enableTLS bool) *NamespaceListerServer {
+	s.useTls = enableTLS
+	return s
+}
+
+func (s *NamespaceListerServer) WithTLSOpts(tlsOpts []func(*tls.Config)) *NamespaceListerServer {
+	s.tlsOpts = tlsOpts
```
filariow, line 89:
```suggestion
	s.tlsOpts = tlsOpts
	s.useTLS = true
```

http_server.go
```diff
@@ -77,6 +80,16 @@ func NewServer(l *slog.Logger, ar authenticator.Request, lister NamespaceLister)
 	}
 }
 
+func (s *NamespaceListerServer) WithTLS(enableTLS bool) *NamespaceListerServer {
+	s.useTls = enableTLS
+	return s
+}
+
```
filariow, line 87:
```suggestion
```

http_server.go
```diff
@@ -92,6 +105,17 @@ func (s *NamespaceListerServer) Start(ctx context.Context) error {
 		}
 	}()
 
+	// setup and serve over TLS if configured
+	if s.useTls {
```
filariow, line 109:
```suggestion
	if s.useTLS {
```

main.go
```diff
@@ -2,12 +2,15 @@ package main
...
 	"context"
+	"crypto/tls"
+	"flag"
 	"log/slog"
 	"os"
 	"os/signal"
 	"syscall"
 
 	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
```
filariow, line 13: Is this right? Shouldn't it be just `errors`?

filariow (changes requested):

main.go
filariow, line 48: Isn't this flag redundant?

filariow (changes requested):

http_server.go
```diff
@@ -77,6 +80,16 @@ func NewServer(l *slog.Logger, ar authenticator.Request, lister NamespaceLister)
 	}
 }
 
+func (s *NamespaceListerServer) WithTLS(enableTLS bool) *NamespaceListerServer {
+	s.useTls = enableTLS
+	return s
+}
+
+func (s *NamespaceListerServer) WithTLSOpts(tlsOpts []func(*tls.Config)) *NamespaceListerServer {
```
filariow, line 88: we may simplify the code in main by using a variadic func here

```suggestion
func (s *NamespaceListerServer) WithTLSOpts(tlsOpts ...func(*tls.Config)) *NamespaceListerServer {
```

main.go
filariow, line 92: we can simplify making WithTLSOpts a variadic func

```suggestion
		tlsOpts := loadTLSCert(l, tlsCertificatePath, tlsCertificateKeyPath)
		s.WithTLSOpts(tlsOpts)
```

main.go
sadlerap: yeah, probably that or just use `fmt.Errorf`, considering that this package is unmaintained.  I blame `gopls` for suggesting this one.

main.go
sadlerap: The convention I've seen on previous projects is to leave these as full sentences.  As an example, look at the options that the workspaces operator takes.

http_server.go
sadlerap: Fixed in f9ea3c6.

main.go
sadlerap: Fixed in f9ea3c6.

main.go
sadlerap: Fixed in f9ea3c6 by making it variadic.

http_server.go
sadlerap: Fixed in f9ea3c6.

http_server.go
sadlerap: Fixed in f9ea3c6.

main.go
sadlerap: Fixed in f9ea3c6.

http_server.go
sadlerap: Fixed in f9ea3c6.

http_server.go
sadlerap: Fixed in f9ea3c6.

main.go
sadlerap: Fixed in f9ea3c6.

http_server.go
```diff
@@ -77,6 +80,12 @@ func NewServer(l *slog.Logger, ar authenticator.Request, lister NamespaceLister)
 	}
 }
 
+func (s *NamespaceListerServer) WithTLSOpts(tlsOpts ...func(*tls.Config)) *NamespaceListerServer {
+	s.tlsOpts = tlsOpts
+	s.useTLS = true
```
filariow, line 85: Formatting is broken here

filariow (changes requested):

main.go
```diff
@@ -21,9 +24,33 @@ func main() {
...
+
 func run(l *slog.Logger) error {
 	log.SetLogger(logr.FromSlogHandler(l.Handler()))
 
+	var enableTLS bool
+	var tlsCertificatePath string
+	var tlsCertificateKeyPath string
+	flag.BoolVar(&enableTLS, "enable-tls", true, "Toggle TLS enablement.")
+	flag.StringVar(&tlsCertificatePath, "cert-path", "", "Path to TLS certificate store.")
+	flag.StringVar(&tlsCertificateKeyPath, "key-path", "", "Path to TLS private key.")
```
filariow, line 51: As `enable-tls` defaults to true, do we need also these two to have defaults?

filariow (commented):
Other than one question, it lgtm

main.go
sadlerap: What would the default be?  I don't know if there's a standard or conventional location to put TLS keys; I've seen `/etc/ssl/`, `/etc/tls/`, `/mnt/`, `/etc/webhook/` (though that seems to be a webhook-specific convention).

http_server.go
sadlerap: Actually, I think I'm going to revert this change and leave `WithTLS` in.  I want to be able to disable TLS, even if someone provides private keys.

http_server.go
sadlerap: Should be fixed as of 7539b6e.

PR #31, opened 2025-01-17
trim cached resources

filariow:
This will reduce the memory usage on big numbers


sadlerap (commented):
Maybe we can also strip annotations?  We never use them for anything, and they can get to be fairly large (i.e. kubectl's last-applied-configuration).  If you want, however, that can be punted to a future PR.

filariow:
@sadlerap done in f892d9c2b62cca8d0be8e3a7592caa6c6df59335

sadlerap (approved):

PR #38, opened 2025-01-24
add performance tests for namespace-lister

filariow:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ ARG TARGETARCH
 WORKDIR /namespace-lister
 
 # Copy the Go Modules manifests
-COPY go.mod go.sum .
+COPY go.mod go.sum ./
 # cache deps before building and copying source so that we don't need to re-download as much
 # and so that source changes don't invalidate our downloaded layer
 RUN go mod download
--- a/Makefile
+++ b/Makefile
@@ -6,8 +6,9 @@ OUT_DIR := $(ROOT_DIR)/out
 GINKGO ?= ginkgo
 GO ?= go
 
+PERF_OUT_DIR := $(OUT_DIR)/perf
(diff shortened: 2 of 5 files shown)


Dockerfile
```diff
@@ -6,7 +6,7 @@ ARG TARGETARCH
 WORKDIR /namespace-lister
 
 # Copy the Go Modules manifests
-COPY go.mod go.sum .
+COPY go.mod go.sum ./
```
sadlerap, line 9: I don't see a difference in behavior - why is this needed?

Makefile
```diff
@@ -3,8 +3,13 @@ LOCALBIN := $(ROOT_DIR)/bin
 
 OUT_DIR := $(ROOT_DIR)/out
 
+GINKGO ?= ginkgo
 GO ?= go
 
+PERF_CLUSTER_PROVIDER ?= kwokctl
+PERF_CLUSTER_KUBECONFIG ?= /tmp/namespace-lister-perf-test
```
sadlerap, line 10: Maybe we should use `mktemp` or use an `out/` directory that we ignore in git?  I'm not a fan of hardcoded paths like this.

Dockerfile
filariow: I applied a fix for a complaint from my linter. When multiple files are copied, it's best practice to explicitly declare a folder.

Makefile
filariow: Great idea

Dockerfile
sadlerap: Then maybe we should have something like this?
```suggestion
COPY go.mod go.sum /namespace-lister/
```

filariow: we are already setting the WORKDIR to `/namespace-lister`. I'd prefer this to remain relative to the WORKDIR

sadlerap (approved):

PR #39, opened 2025-01-24
expose metrics on /metrics

sadlerap:
We expose three metrics:
- api_counter, which counts the number of requests we receive.  It's broken down by the method and response code of each request we process.
- api_latency, which measures the latency of each request we process. This is reported as a histogram.
- api_response_size, which measures how large each response we send is. This is also reported as a histogram.


filariow (approved):
/lgtm

Maybe also https://pkg.go.dev/github.com/prometheus/client_golang/prometheus/promhttp#InstrumentHandlerInFlight

filariow (approved):
/lgtm

well done! 🎉

PR #46, opened 2025-01-30
introduce accesses caching

filariow:


pkg/auth/cache/auth_cache.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"sync/atomic"
+
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+)
+
+// stores data
+type AuthCache struct {
+	data atomic.Pointer[map[rbacv1.Subject][]corev1.Namespace]
```
sadlerap, line 12: I'm thinking we may need a `sync.RWMutex` instead of an atomic pointer swap to synchronize access, but I'm not 100% sure.

sadlerap (commented):
A few comments, otherwise looks fine

Makefile
```diff
@@ -61,6 +61,7 @@ test-perf: ## Run performance tests
 		--name namespace-lister-perf-test $(PERF_CLUSTER_PROVIDER_FLAGS)
 	KUBECONFIG=$(PERF_CLUSTER_KUBECONFIG) $(GINKGO) --label-filter='perf' \
 		--keep-going --procs=1 --flake-attempts 2 --output-dir=$(PERF_OUT_DIR)
+	-$(PERF_CLUSTER_PROVIDER) delete cluster --name namespace-lister-perf-test
```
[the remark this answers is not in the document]
filariow: yeah, I added this as I always forgot it running. Let me remove that line for this PR

pkg/auth/cache/auth_cache.go
filariow: until access is readonly we should be fine without a mutex, right?

sadlerap: Correct, but long-term I think it's what we should probably be using to synchronize accesses.

sadlerap (approved):
lgtm, I think we're safe to try to deploy/test this.  We can always revert if we have issues.

PR #54, opened 2025-02-04
deploy metrics as a separate service

sadlerap:


config/deployment.yaml
```diff
@@ -23,6 +23,7 @@ spec:
         - -enable-tls
         - -cert-path=/var/tls/tls.crt
         - -key-path=/var/tls/tls.key
+        - -metrics-address=0.0.0.0:9443
```
filariow, line 26: I'd prefer to reserve port 9443 for the API server

main.go
```diff
@@ -89,9 +91,22 @@ func run(l *slog.Logger) error {
...
+	reg := NewDefaultRegistry()
+	metrics := NewMetrics(reg)
+
+	// serve metrics
+	metricsServer := NewMetricsServerBuilder().
+		WithAddress(metricsListenAddress).
+		WithTLS(enableTLS).
+		WithTLSConfig(loadTLSCert(l, tlsCertificatePath, tlsCertificateKeyPath)).
+		WithRegistry(reg).
+		Build()
```
filariow, line 104: I'd suggest to provide the required parameters in the constructor and leave the `With*` approach for optional ones

main.go
```diff
@@ -89,9 +91,22 @@ func run(l *slog.Logger) error {
...
+	go metricsServer.Serve(ctx)
```
filariow, line 105: if the metrics server terminates, we may still want to call `cancel()` to be sure sure `ctx` is invalidated

metrics_server.go
```diff
@@ -0,0 +1,110 @@
...
+	} else {
+		logger.Info("Serving metrics over http", "address", ms.Server.Addr)
+		err = ms.Server.ListenAndServe()
+	}
+
+	if errors.Is(err, http.ErrServerClosed) {
+		logger.Info("gracefully shutting down metrics server")
+	} else {
+		logger.Error("metrics server closed unexpectedly", "error", err)
+	}
```
filariow, line 109: I find if/else hard to read. Can we move these blocks to auxiliary functions?

```go
if condition {
    return doTrue()
}
return doFalse()
```

filariow (changes requested):
some nits

filariow:
superseded by #51

PR #55, opened 2025-02-04
yamllint: increase allowed line length

sadlerap:
We've been getting a warning about a too-long line in ci.yaml:149.  I wasn't able to come up with a way of splitting this line that was pleasing to my eyes, so increase the allowed line length instead.


filariow (approved):
/lgtm

PR #58, opened 2025-02-07
🐛 remove duplicate namespaces from server responses

filariow:
If access is provided via multiple RoleBindings/ClusterRoleBindings, or via a combination of them, the same namespace can be present more than once in server responses.
This PR introduces some unit tests for this scenario and a patch.

- **add failing tests**
- **remove duplicates from cache**


sadlerap (approved):
Looks fine to me.

I think there may be some ways to make things go faster & still be correct, such as using sets instead of slices.

PR #59, opened 2025-02-07
add performance test for cache Synch

filariow:

### DIFF
--- a/performance_test.go
+++ b/performance_test.go
@@ -246,6 +246,44 @@ var _ = Describe("Authorizing requests", Serial, Ordered, func() {
 		}, gmeasure.SamplingConfig{N: 30, Duration: 2 * time.Minute})
 		// we'll sample the function up to 30 times or up to 2 minutes, whichever comes first.
 	})
+
+	It("efficiently synchronizes access cache", Serial, Label("perf"), func(ctx context.Context) {
+		// new gomega experiment
+		experiment := gmeasure.NewExperiment("Access Cache Synch")
+
+		// Register the experiment as a ReportEntry - this will cause Ginkgo's reporter infrastructure
+		// to print out the experiment's report and to include the experiment in any generated reports
+		AddReportEntry(experiment.Name, experiment)
+
+		// create cache, namespacelister, and handler
+		cache, err := BuildAndStartResourceCache(ctx, cacheCfg)
+		utilruntime.Must(err)
+		c, err := buildAndStartAccessCache(ctx, cache)
+		utilruntime.Must(err)
+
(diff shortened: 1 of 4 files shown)


performance_test.go
```diff
@@ -246,8 +251,80 @@ var _ = Describe("Authorizing requests", Serial, Ordered, func() {
...
+		// we get the median synch duration from the experiment we just ran
+		httpListingStats := experiment.GetStats("cache synch")
+		medianDuration := httpListingStats.DurationFor(gmeasure.StatMedian)
+
+		// and assert that it is below a threshold
+		Expect(medianDuration).To(BeNumerically("<=", 200*time.Millisecond))
+	})
 })
 
+// unsafeGetPrivateCacheData retrieves the map used by the cache to store data.
```
filariow, line 307: @sadlerap what do you think about this? It's really unsafe, but the only alternative I was able to think was to expose some methods from the AccessCache that were only used in tests, and I'm not a fan of it...

performance_test.go
```diff
@@ -246,8 +251,80 @@ var _ = Describe("Authorizing requests", Serial, Ordered, func() {
...
+		utilruntime.Must(err)
+		c, err := buildAndStartAccessCache(ctx, resourceCache)
+		utilruntime.Must(err)
+
+		// check cache is correctly populated with
+		// more than 5000 subjects
+		// and more than 10000 total namespaces
+		cacheData := unsafeGetPrivateCacheData(c.AccessCache)
+		Expect(len(cacheData)).To(BeNumerically(">", 5000))
+		Expect(cacheData).To(Satisfy(func(d cache.AccessData) bool {
```
sadlerap, line 274: Shouldn't this closure operate on a `map[rbacv1.Subject][]corev1.Namespace`?

performance_test.go
sadlerap: The short-term solution could be to provide method giving access to the map.  Make it a private method, since we don't want anyone outside of the `cache` package to ever call it.  This benchmark would also need to be moved into that package.

The longer-term solution would be to expose this information from the cache directly and assert on it instead.  We could even hook it up to metrics to get an idea of how big the cache is in our observability stack.

performance_test.go
filariow: This way if we make changes to what AccessCache uses, it will fail at compile time

performance_test.go
filariow: I like the idea of instrumenting the cache and taking info from there.

I'd propose to add a story to our backlog for future implementation and to now rely on this `unsafe` approach. Is it ok for you?

sadlerap (approved):

PR #62, opened 2025-02-07
cache: store namespace names instead of metadata

sadlerap:
Instead of caching namespace metadata, we can instead only cache the namespaces' names and pull their definitions from the client cache. This saves some memory at the expense of a minor increase in response latency (benchmarks indicated on the order of a few extra milliseconds).

### DIFF
--- a/pkg/auth/cache/synchronized_access_cache.go
+++ b/pkg/auth/cache/synchronized_access_cache.go
@@ -12,6 +12,8 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apiserver/pkg/authorization/authorizer"
 	"k8s.io/kubernetes/plugin/pkg/auth/authorizer/rbac"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -25,7 +27,7 @@ func isSynchAlreadyRunningErr(err error) bool {
 
 // applies changes to cache async
 type SynchronizedAccessCache struct {
-	*AccessCache
+	cache         *AccessCache
 	request       chan struct{}
 	synchronizing atomic.Bool
 	once          sync.Once
@@ -44,8 +46,8 @@ func NewSynchronizedAccessCache(
 	opts CacheSynchronizerOptions,
 ) *SynchronizedAccessCache {
 	return opts.Apply(&SynchronizedAccessCache{
-		AccessCache: NewAccessCache(),
-		request:     make(chan struct{}, 1),
+		cache:   NewAccessCache(),
+		request: make(chan struct{}, 1),
 
 		subjectLocator:  subjectLocator,
 		namespaceLister: namespaceLister,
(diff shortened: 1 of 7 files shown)


pkg/auth/cache/synchronized_access_cache.go
```diff
@@ -91,12 +93,16 @@ func (s *SynchronizedAccessCache) Synch(ctx context.Context) error {
 
 		// store in temp cache
 		for _, s := range ss {
-			c[s] = append(c[s], ns)
+			if _, exists := c[s]; !exists {
+				c[s] = sets.New(ns.Name)
+			} else {
+				c[s].Insert(ns.Name)
+			}
```
filariow, line 100: if we rely on sets here, we can remove the `removeDuplicateSubjects` function and its invocation at L92

pkg/auth/cache/synchronized_access_cache.go
```diff
@@ -91,12 +93,16 @@ func (s *SynchronizedAccessCache) Synch(ctx context.Context) error {
 
 		// store in temp cache
 		for _, s := range ss {
-			c[s] = append(c[s], ns)
+			if _, exists := c[s]; !exists {
+				c[s] = sets.New(ns.Name)
+			} else {
```
filariow, line 98: do you think we can remove this `else` somehow?

pkg/auth/cache/synchronized_access_cache.go
```diff
@@ -179,3 +185,21 @@ func (s *SynchronizedAccessCache) Start(ctx context.Context) {
...
 	})
 }
+
+func (s *SynchronizedAccessCache) List(ctx context.Context, subject rbacv1.Subject) []corev1.Namespace {
+	namespaces := s.cache.List(subject)
+	namespaceList := make([]corev1.Namespace, 0, len(namespaces))
+
+	for _, ns := range namespaces {
+		namespace := corev1.Namespace{}
+		err := s.namespaceLister.Get(ctx, types.NamespacedName{Name: ns}, &namespace)
```
filariow, line 195: this way we are getting access from the `AccessCache` and namespaces from the `ResourcesCache`. This is coupling the `List` operation to the two caches.

It may happen that the namespaces in the `ResourcesCache` were updated before the access in the `AccessCache` and this may introduce some errors.
I'd suggest to cache the namespaces used during the synch operation in the `AccessCache`, so that we always provide consistent results.

filariow: Also if we cache it in the AccessCache, we can implement a slightly more performant version of this algorithm by sorting namespaces ahead-of-time and using binary search just-in-time

pkg/auth/cache/synchronized_access_cache.go
```diff
@@ -179,3 +185,21 @@ func (s *SynchronizedAccessCache) Start(ctx context.Context) {
 		}()
 	})
 }
+
+func (s *SynchronizedAccessCache) List(ctx context.Context, subject rbacv1.Subject) []corev1.Namespace {
```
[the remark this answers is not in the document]
sadlerap: I don't agree.  I think the `AccessCache` should solely be concerned with storage. `SynchronizedAccessCache`, meanwhile, should be able to provide the additional features we need in our cache, relying on `AccessCache` to provide synchronization.

In other words: I don't see the benefit of pushing this down a layer.

PR #79, opened 2025-02-20
feat: instrument access cache

filariow:
This PR instruments AccessCache so that we can collect metrics.


pkg/auth/cache/metrics_accesscache_test.go
```diff
@@ -0,0 +1,227 @@
...
+)
+
+const (
+	resourcesRequestsMetricFullname     = "namespace_lister_accesscache_resource_requests_total"
+	timeRequestsMetricFullname          = "namespace_lister_accesscache_time_requests_total"
+	syncMetricFullname                  = "namespace_lister_accesscache_synch_op_total"
+	subjectNamespacePairsMetricFullname = "namespace_lister_accesscache_subject_namespace_pairs"
+	subjectsMetricFullname              = "namespace_lister_accesscache_subjects"
+
+	metadataSynchOp = `# HELP namespace_lister_accesscache_synch_op_total synchronization operations
```
sadlerap, line 29: We should maybe use [prometheus's common library](https://pkg.go.dev/github.com/prometheus/common@v0.62.0/expfmt) to parse this data, rather than doing string comparisons.  We can do that in a follow-up PR if you want.

sadlerap (commented):
Just a thought, otherwise LGTM

pkg/auth/cache/metrics_accesscache_test.go
filariow: Good ideas. Let's look into that on a follow up PR

sadlerap (approved):

PR #91, opened 2025-03-03
metrics: expose number of cached namespaces per subject

sadlerap:
This series exposes the number of namespaces a subject has in our cache.  This allows us to have better metrics on how our cache is being used in live deployments.

By adding these metrics, we can add more insightful metrics to our dashboards.  For instance, we can now tell if there are subjects in our cache who have a lot of cached namespaces, indicating broad access across the cluster.  We can also use these metrics to get an idea of how large our cache is, by summing across all metrics and ignoring labels.

This series also does a little refactoring of the unit tests associated with the cache.

### DIFF
--- a/pkg/auth/cache/metrics_accesscache.go
+++ b/pkg/auth/cache/metrics_accesscache.go
@@ -1,7 +1,10 @@
 package cache
 
 import (
+	"fmt"
+
 	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/api/rbac/v1alpha1"
 )
 
 const (
@@ -41,6 +44,9 @@ type accessCacheMetrics struct {
 	// synchGauge counts the number of cache synchronization
 	synchGauge *prometheus.CounterVec
 
+	// subjectCachedNamespaceCounter counts how many namespaces a user has in the cache.
+	subjectCachedNamespaceCounter *prometheus.GaugeVec
+
 	// resourceRequestsGauge counts the number of cache synchronization
 	// requested as a consequence of resource events
 	resourceRequestsGauge *prometheus.CounterVec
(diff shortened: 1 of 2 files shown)


pkg/auth/cache/metrics_accesscache.go
```diff
@@ -73,6 +76,12 @@ func NewAccessCacheMetrics() AccessCacheMetrics {
 			"status",
 			"error",
 		}),
+		subjectCachedNamespaceCounter: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Namespace: "namespace_lister",
+			Subsystem: "accesscache",
+			Name:      "cached_namespace_count",
+			Help:      "number of cached namespaces",
+		}, []string{"apiGroup", "kind", "name", "namespace"}),
```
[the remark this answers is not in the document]
sadlerap: Fixed in df8f2bb.

pkg/auth/cache/metrics_accesscache.go
```diff
@@ -154,10 +170,26 @@ func (s *accessCacheMetrics) CollectSynchMetrics(cacheData AccessData, err error
...
+
 	// update number of (subject, namespace) pairs
 	subNsPairCount := 0
-	for _, v := range cacheData {
+	for k, v := range cacheData {
 		subNsPairCount += len(v)
+
+		var user string
+		if k.Kind == v1alpha1.ServiceAccountKind {
+			user = fmt.Sprintf("system:serviceaccount:%s:%s", k.Name, k.Namespace)
```
filariow, line 184:
```suggestion
			user = fmt.Sprintf("system:serviceaccount:%s:%s", k.Namespace, k.Name)
```

pkg/auth/cache/metrics_accesscache.go
```diff
@@ -73,6 +79,12 @@ func NewAccessCacheMetrics() AccessCacheMetrics {
 			"status",
 			"error",
 		}),
+		subjectCachedNamespaceCounter: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Namespace: "namespace_lister",
+			Subsystem: "accesscache",
+			Name:      "cached_namespace_count",
+			Help:      "number of cached namespaces",
+		}, []string{"apiGroup", "user"}),
```
filariow, line 87:
```suggestion
		}, []string{"apiGroup", "subject"}),
```

Wouldn't it be more appropriate to have subject instead of user here?

filariow:
I'm concerned about the amount of data we are going to store with this label

Ref https://prometheus.io/docs/practices/naming/

> CAUTION: Remember that every unique combination of key-value label pairs represents a new time series, which can dramatically increase the amount of data stored. Do not use labels to store dimensions with high cardinality (many different label values), such as user IDs, email addresses, or other unbounded sets of values.

filariow:
@sadlerap do you think we can close this one?

PR #94, opened 2025-03-10
add egress networkpolicy

filariow:
this will restrict the namespace-lister to only speak with the APIServer


sadlerap (approved):
The label selectors for api-server will be need to changed when we roll this out to infra-deployments, but we can address that there.

PR #95, opened 2025-03-10
add healthz and readyz endpoints

sadlerap:
Add /healthz and /readyz endpoints as probes for kubernetes to talk to.

For now, readyz and healthz do the same thing, but we may want to have readyz have more checks in the future, such as ensuring we can talk to apiserver and we have sufficient privileges to run.


filariow (approved):
/lgtm

PR #101, opened 2025-03-17
Add group support in access cache

filariow:


sadlerap (approved):

http_api_handler_list.go
```diff
@@ -29,7 +29,7 @@ func (h *ListNamespacesHandler) ServeHTTP(w http.ResponseWriter, r *http.Request
 	ud := r.Context().Value(ContextKeyUserDetails).(*authenticator.Response)
 
 	// retrieve projects as the user
-	nn, err := h.lister.ListNamespaces(r.Context(), ud.User.GetName())
+	nn, err := h.lister.ListNamespaces(r.Context(), ud.User.GetName(), nil)
```
gbenhaim, line 32: 1. from where it will get the list of groups? is it from the http headers?
2. the system:authenticated group is a built in group. it won't be included as part of the headers in the http request.

filariow: cf #102, #106, #107

1. in the smart-proxy configuration, the namespace-lister expects the groups to be provided by the proxy through an header. The proxy is in charge of setting the header accordingly then
1. in the dumb-proxy configuration, the information is returned by the TokenAccessReview API

gbenhaim: so we need to hard code the inclusion of the `system:authenticated` group? (for the smart proxy case)

filariow: we need the proxy to add the `Impersonate-Group: system:authenticated` header if the request was authenticated correctly. Ideally, it should add `Impersonate-Group: system:anonymous` otherwise.

filariow: ATM, the namespace-lister is not supporting the anonymous scenario though

PR #102, opened 2025-03-17
add groups support to the HTTP layer

filariow:

### DIFF
--- a/authenticator.go
+++ b/authenticator.go
@@ -27,6 +27,7 @@ import (
 // APIServer replies are cached for a short time.
 type Authenticator struct {
 	usernameHeader string
+	groupsHeader   string
 	next           authenticator.Request
 }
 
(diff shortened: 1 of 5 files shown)


authenticator.go
```diff
@@ -104,7 +108,13 @@ func newTokenReviewAuthenticatorWithOpts(opts *AuthenticatorOptions) (authentica
...
+// to use when authenticating requests by username header to extract user's username
 func GetUsernameHeaderFromEnv() string {
 	return os.Getenv(EnvUsernameHeader)
 }
+
+// GetGroupsHeaderFromEnv retrieves from environment variable the name of the header
+// to use when authenticating requests by username header to extract user's groups
+func GetGroupsHeaderFromEnv() string {
+	return os.Getenv(EnvGroupsHeader)
+}
```
sadlerap, line 120: We should probably test group membership in e2e tests somehow, maybe against something simple like `system:authenticated` or similar.  I'd like to see what changes we need to make to the nginx configuration to pass on group membership.

filariow: That's right. I'm working at them in a separate PR

sadlerap (approved):

PR #114, opened 2025-03-26
add synch timeout

filariow:
This PR introduces a timeout to the synch operation.

It can be set from configuration. If not, the maximum between 1min and `resyncPeriod-1min` is used.

### DIFF
--- a/pkg/auth/cache/synchronized_access_cache_test.go
+++ b/pkg/auth/cache/synchronized_access_cache_test.go
@@ -11,6 +11,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apiserver/pkg/authorization/authorizer"
 	"k8s.io/kubernetes/plugin/pkg/auth/authorizer/rbac"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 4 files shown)


pkg/auth/cache/synchronized_access_cache_test.go
```diff
@@ -57,6 +58,68 @@ var _ = Describe("SynchronizedAccessCache", func() {
...
+			DoAndReturn(func(attributes authorizer.Attributes) ([]rbacv1.Subject, error) {
+				if !hasBeenCalledOnce {
+					hasBeenCalledOnce = true
+					return []rbacv1.Subject{userSubject}, nil
+				}
+
+				time.Sleep(100 * time.Millisecond)
+				return []rbacv1.Subject{userSubject}, nil
+			}).
+			Times(2)
```
sadlerap, line 111: We can clean this up to not need the `hasBeenCalledOnce` bookkeeping:

<details><summary>Diff</summary>
<p>

```diff
diff --git a/pkg/auth/cache/synchronized_access_cache_test.go b/pkg/auth/cache/synchronized_access_cache_test.go
index f56d85f..5ead6fb 100644
--- a/pkg/auth/cache/synchronized_access_cache_test.go
+++ b/pkg/auth/cache/synchronized_access_cache_test.go
@@ -95,20 +95,17 @@ var _ = Describe("SynchronizedAccessCache", func() {
 			}).
 			Times(1)
 
-			// subjectLocator will reply fast the first time it's called and cause a timeout the second time
-		hasBeenCalledOnce := false
 		subjectLocator.EXPECT().
 			AllowedSubjects(gomock.Any()).
 			DoAndReturn(func(attributes authorizer.Attributes) ([]rbacv1.Subject, error) {
-				if !hasBeenCalledOnce {
-					hasBeenCalledOnce = true
-					return []rbacv1.Subject{userSubject}, nil
-				}
-
-				time.Sleep(100 * time.Millisecond)
+				// reply fast the first time
 				return []rbacv1.Subject{userSubject}, nil
 			}).
-			Times(2)
+			DoAndReturn(func(attributes authorizer.Attributes) ([]rbacv1.Subject, error) {
+				// cause a timeout the second time
+				time.Sleep(100 * time.Millisecond)
+				return []rbacv1.Subject{userSubject}, nil
+			})
 
 		nsc := cache.NewSynchronizedAccessCache(subjectLocator, namespaceLister, cache.CacheSynchronizerOptions{
 			SynchTimeout: 50 * time.Millisecond,

```

</p>
</details>

sadlerap (commented):
other than one recommendation, lgtm

pkg/auth/cache/synchronized_access_cache_test.go
filariow: oh, that's nice! we still need the times 2, though. Right?

sadlerap: Nope, it'll execute each sequentially.  We had only two returns the first time, so we need only two returns here.

filariow: 1ba5c88e0cf8c5eeab8389cbde92721f5bedf30b

pkg/auth/cache/synchronized_access_cache_test.go
```diff
@@ -57,6 +58,65 @@ var _ = Describe("SynchronizedAccessCache", func() {
...
+				return nil
+			}).
+			Times(1)
+
+		subjectLocator.EXPECT().
+			AllowedSubjects(gomock.Any()).
+			DoAndReturn(func(attributes authorizer.Attributes) ([]rbacv1.Subject, error) {
+				// reply fast the first time
+				return []rbacv1.Subject{userSubject}, nil
+			}).
```
sadlerap, line 103: Okay, now that I read my own suggestion, there's an even better way to do this one:
```suggestion
			// reply fast the first time
			Return([]rbacv1.Subject{userSubject}, nil).
```

I think my brain must be malfunctioning today :laughing:

sadlerap (approved):
with this, lgtm

PR #139, opened 2025-04-25
ci: pull go version from go.mod

sadlerap:
We can tell the setup-go action to pull the go version directly from go.mod.  Use that rather than doing awkward go.mod parsing ourselves.


filariow (approved):
Nice!

/lgtm

PR #140, opened 2025-04-25
ci: use golangci-lint v2

sadlerap:
The selector we were using for golangci-lint's version wasn't picking up golangci-lint's bump to v2 a while back.  Specify v2 when doing version selection.


filariow (approved):
/lgtm

PR #333, opened 2026-04-17
KFLUXINFRA-3243 - Add additional e2e tests

oswcab:


acceptance/make/common.mk
```diff
@@ -50,6 +54,16 @@ deploy-test-infra:
 		--timeout=300s \
 		-l 'app.kubernetes.io/instance=cert-manager' \
 		-n cert-manager deployment
+	@echo "Waiting for cert-manager webhook to accept connections..."
+	@for i in $$(seq 1 30); do \
+		if $(KUBECTL) get validatingwebhookconfigurations cert-manager-webhook >/dev/null 2>&1 && \
+		   $(KUBECTL) apply --dry-run=server -f - >/dev/null 2>&1 <<< \
+		   '{"apiVersion":"cert-manager.io/v1","kind":"Issuer","metadata":{"name":"probe","namespace":"default"},"spec":{"selfSigned":{}}}'; then \
+			break; \
```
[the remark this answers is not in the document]
oswcab: Done.

oswcab:
> ## PR Reviewer Guide 🔍
> #### (Review updated until commit [fe59a33](https://github.com/konflux-ci/namespace-lister/commit/fe59a33fd5a69098e79d330e8e4704a7ea779206))
> Here are some key observations to aid the review process:
> 
> ⏱️ **Estimated effort to review**: 3 🔵🔵🔵⚪⚪
> 🧪 **PR contains tests**
> 🔒 **Security concerns**
> 
> **TLS verification disabled:** `InsecureSkipVerify: true` is set in acceptance test hooks (dumb-proxy and smart-proxy). While likely acceptable for tests, it weakens TLS security; ensure it is not used outside tests and that CI environments are not unintentionally trusting MITM connections.
> 
> ⚡ **Recommended focus areas for review**
> 
> [**Compile Error**](https://github.com/konflux-ci/namespace-lister/pull/333/files#diff-6dcda7086e0bc4cf35395c260ca69a653dc63f2397bb4c89de3a6e2b87335183R212-R249)
> The bulk namespace creation loop ranges over an `int` (`for i := range toCreate`), which is not valid Go and will fail compilation. It should iterate with a conventional counter loop to create `toCreate` namespaces.
> 
> ```go
> // create new namespaces
> tn := time.Now().Unix()
> for i := range toCreate {
> 	name := fmt.Sprintf("run-%s-bulk-%d-%d", run, tn, i)
> 	ns := &corev1.Namespace{
> 		ObjectMeta: metav1.ObjectMeta{
> 			Name: name,
> 			Labels: map[string]string{
> 				"konflux.ci/type":           "user",
> 				"namespace-lister/scope":    "acceptance-tests",
> 				"namespace-lister/test-run": run,
> 			},
> 		},
> 	}
> 	if err := cli.Create(ctx, ns); err != nil {
> 		return ctx, err
> 	}
> 
> 	rb := &rbacv1.RoleBinding{
> 		ObjectMeta: metav1.ObjectMeta{
> 			Name:      fmt.Sprintf("run-%s-bulk-%d-%d-access", run, tn, i),
> 			Namespace: name,
> 		},
> 		RoleRef: rbacv1.RoleRef{
> 			Kind:     "ClusterRole",
> 			Name:     "namespace-get",
> 			APIGroup: rbacv1.GroupName,
> 		},
> 		Subjects: []rbacv1.Subject{user.AsSubject()},
> 	}
> 	if err := cli.Create(ctx, rb); err != nil {
> 		return ctx, err
> 	}
> 
> 	nn = append(nn, *ns)
> }
> 
> return tcontext.WithNamespaces(ctx, nn), nil
> ```
> 
> [**TLS Handling**](https://github.com/konflux-ci/namespace-lister/pull/333/files#diff-08df89e5ea4d9611c2be829685b52e5c8d23b6b4d2c4535a448181b3116d5465R28-R34)
> The tests set `InsecureSkipVerify: true` for TLS. Validate that this is strictly limited to test code paths and cannot leak into production-like configs, and ensure the address scheme/CA expectations are correct for the environment.
> 
> ```go
> func injectServiceAddresses(ctx context.Context, sc *godog.Scenario) (context.Context, error) {
> 	address := suite.EnvKonfluxAddressOrDefault(defaultTestAddress)
> 	ctx = tcontext.WithNamespaceListerAddress(ctx, address)
> 	ctx = tcontext.WithMetricsAddress(ctx, address)
> 	ctx = tcontext.WithTLSConfig(ctx, &tls.Config{InsecureSkipVerify: true}) //nolint:gosec
> 	return ctx, nil
> }
> ```
> 
> [**Data Mutability**](https://github.com/konflux-ci/namespace-lister/pull/333/files#diff-9601977a092ae2d4047b7c3a825d138b21ead9f6a53494b740d90b48ca73d991R82-R94)
> `WithHTTPResponse` stores a `[]byte` directly in context. If the passed slice is later mutated by the caller, the stored response body changes unexpectedly. Consider copying the slice before storing to keep the context value immutable.
> 
> ```go
> func WithHTTPResponse(ctx context.Context, statusCode int, body []byte) context.Context {
> 	resp := HTTPResponse{StatusCode: statusCode, Body: body}
> 	return into(ctx, ContextKeyHTTPResponse, resp)
> }
> 
> func GetHTTPResponse(ctx context.Context) HTTPResponse {
> 	return get[HTTPResponse](ctx, ContextKeyHTTPResponse)
> }
> 
> type HTTPResponse struct {
> 	StatusCode int
> 	Body       []byte
> }
> ```

1. "Compile Error" with `for i := range toCreate` -- Invalid finding. Integer range (`for i :=  range n`) has been valid Go since Go 1.22 (spec). The project uses Go 1.25+ and we verified it compiles with go vet ./.... 

2. TLS InsecureSkipVerify -- Low concern. It's in test hook files (`hook_test.go`) inside `_test.go` packages, only injected into the Godog scenario context for Kind clusters with self-signed certs. It can't leak into production code -- these files aren't part of the main module and are only compiled by go test. The `//nolint:gosec` annotation already acknowledges
the intent. This is intentional for test environment with self-signed certs.

3. `WithHTTPResponse` storing []byte directly -- Technically valid but not a real concern here. The only caller is `metricsEndpointReturnsPrometheusFormat` in `steps_metrics.go`, which stores the body from `io.ReadAll` and never mutates it afterward. The next step (`metricsContainMetricName`) only reads it and there's no mutation path. That said, a defensive copy is cheap: => changed.

oswcab:
> The cert-manager webhook readiness probe relies on a specific `validatingwebhookconfigurations` name and a server-side dry-run apply; this can still be timing-sensitive and may be brittle across cert-manager versions or cluster states. Validate that this logic is stable in CI and won’t introduce intermittent failures.

The approach is already the most robust option available:
  - The webhook name `cert-manager-webhook` is stable across `cert-manager` versions (it's part of their public API)
  - The dry-run probes the actual webhook path, not just the deployment status
  - The 30-retry loop with 2s intervals gives 60s of tolerance
  - We fail explicitly if it never becomes ready

The alternative would be a simple sleep 30 which is less reliable and slower. The probe tests the actual webhook endpoint rather than assuming readiness from deployment status. The webhook configuration name is stable across `cert-manager` versions. The 60s retry window has been sufficient in local and CI testing.

.github/workflows/ci.yaml
```diff
@@ -107,10 +107,18 @@ jobs:
...
           echo "##[endgroup]"
 
+      - name: Upload test results
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: acceptance-test-results-${{ matrix.proxy }}
+          path: acceptance/test/${{ matrix.proxy }}/out/
+          retention-days: 7
+
```
filariow, line 121: what is this needed for?

acceptance/dependencies/cert-manager/kustomization.yml
```diff
@@ -3,44 +3,3 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
   - https://github.com/cert-manager/cert-manager/releases/download/v1.14.4/cert-manager.yaml
-
```
filariow, line 6: why are we removing these patches?

acceptance/config/patches/with_cache_resyncperiod.yaml
```diff
@@ -2,4 +2,4 @@
   path: /spec/template/spec/containers/0/env/-
   value:
     name: CACHE_RESYNC_PERIOD
-    value: '20s'
+    value: '5s'
```
filariow: I'm not sure this is needed. The resync is triggered by any event on watched Namespaces, Roles, RoleBindings, ClusterRoles, and ClusterRoleBindings. If nothing happens for more than `CACHE_RESYNC_PERIOD` we run a new resync. I think we can revert this change

acceptance/make/common.mk
```diff
@@ -31,12 +31,16 @@ image-build:
...
+	-$(KIND) delete cluster --name "$(KIND_CLUSTER_NAME)"
 	$(KIND) create cluster --name "$(KIND_CLUSTER_NAME)" --config kind-config.yaml
+	$(KUBECTL) wait --for=condition=Ready node --all --timeout=120s
 
 .PHONY: kind-load-image
-kind-load-image:
-	$(IMAGE_BUILDER) save $(IMG) | \
-		$(KIND) load image-archive --name $(KIND_CLUSTER_NAME) /dev/stdin
+kind-load-image: $(OUT_DIR)
+	$(IMAGE_BUILDER) tag $(IMG) docker.io/library/$(IMG) 2>/dev/null || true
```
filariow: why is this needed?

acceptance/make/common.mk
```diff
@@ -31,12 +31,16 @@ image-build:
...
+	$(IMAGE_BUILDER) save docker.io/library/$(IMG) -o $(OUT_DIR)/image.tar
+	$(KIND) load image-archive $(OUT_DIR)/image.tar --name $(KIND_CLUSTER_NAME)
+	rm -f $(OUT_DIR)/image.tar
```
filariow: Any reason not to pipe through `/dev/stdin` anymore?

acceptance/dependencies/cert-manager/kustomization.yml
```diff
@@ -3,44 +3,3 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
   - https://github.com/cert-manager/cert-manager/releases/download/v1.14.4/cert-manager.yaml
```
filariow, line 5: it's quite old. we can update it. Not sure if it's in the scope of this PR though 😄

```suggestion
  - https://github.com/cert-manager/cert-manager/releases/download/v1.20.2/cert-manager.yaml
```

acceptance/make/common.mk
```diff
@@ -50,6 +54,22 @@ deploy-test-infra:
...
+			webhook_ready=true; \
+			break; \
+		fi; \
+		echo "  webhook not ready yet, retrying ($$i/30)..."; \
+		sleep 2; \
+	done; \
+	if [ "$$webhook_ready" != "true" ]; then \
+		echo "ERROR: cert-manager webhook did not become ready after 30 attempts"; \
+		exit 1; \
+	fi
```
filariow, line 68: mixed feelings on this: I find it complex and a little cumbersome. Why do we need this? are we seeing a good amount of flakiness coming from this?

I would expect a complete rollout to be enough, ie. (Mutating/Validating)WebhookConfig to be there and readiness checks to pass on cert-manager's pods. 

What do you think about setting `pods.spec.containers.readinessProbe.successThreshold` to an higher value instead? like `3` or `5`?

filariow (changes requested):
I'm half-way. Leaving some initial feedback. It would be great if we could split this one in multiple PRs.

acceptance/make/common.mk
```diff
@@ -31,12 +31,16 @@ image-build:
 
 .PHONY: kind-create
 kind-create:
+	-$(KIND) delete cluster --name "$(KIND_CLUSTER_NAME)"
```
filariow: this is changing a bit the semantic of the make target (`kind-create`). It's moving it from `create` to something like `ensure-clean-cluster-exists`. 
If we don't need this logic, I'd suggest to revert this change. If do we need it, I'd suggest to create a new target 😄

acceptance/make/common.mk
```diff
@@ -31,12 +31,16 @@ image-build:
...
 	$(KIND) create cluster --name "$(KIND_CLUSTER_NAME)" --config kind-config.yaml
+	$(KUBECTL) wait --for=condition=Ready node --all --timeout=120s
```
filariow: Is this needed? The node usually becomes Ready a little later (in the order of a few tens of seconds), but the APIServer is already working. I feel like this will slow us down a bit with no real gain.

.github/workflows/ci.yaml
oswcab: This uploads test output (logs, reports) as artifacts so we can inspect failures after the workflow completes, especially useful for flaky or environment-dependent issues where re-reading CI logs alone isn't enough. The `if: always()` ensures we get artifacts even on failure, and they auto-expire after 7 days.

acceptance/dependencies/cert-manager/kustomization.yml
oswcab: This change is no longer part of this PR, it is in the infrastructure fixes PR ([#k3243-pre](https://github.com/konflux-ci/namespace-lister/pull/337)). The limits (90m CPU / 120Mi memory) were causing CrashLoopBackOff on `cert-manager-cainjector` and `webhook` as they're too tight for startup when running the tests locally. Removing them lets `cert-manager` use its own defaults, which are already reasonable for a test cluster.

acceptance/config/patches/with_cache_resyncperiod.yaml
oswcab: You're right — the cache tests already poll with Eventually, so they don't depend on a short resync period. I'll revert this to 20s.

acceptance/make/common.mk
oswcab: Podman stores locally-built images as `localhost/namespace-lister:latest`, but containerd (inside Kind nodes) resolves unqualified names as docker.io/library/.... Without the explicit tag, the image loads into Kind but pods fail with `ImagePullBackOff` because the names don't match. This makes make test-e2e work with `podman` out of the box.

acceptance/make/common.mk
oswcab: `podman save` + pipe to kind load image-archive /dev/stdin was unreliable in some podman versions (hangs or broken pipe). Saving to a temp file first is more portable across `Docker` and `podman`. Happy to add a comment explaining this if it helps.

acceptance/dependencies/cert-manager/kustomization.yml
oswcab: Good idea, I'll update it in this PR since we're already touching this file.

acceptance/make/common.mk
oswcab: The issue we hit wasn't pod readiness — it was the webhook admission endpoint not accepting connections yet (even after pods reported Ready). The dry-run kubectl apply is the only reliable way to confirm the webhook is actually serving. That said, I could simplify it: reduce retries from 30 to 15 and increase sleep to 3s, since the webhook usually comes up within 20-30s. Would that be acceptable?

acceptance/make/common.mk
oswcab: Makes sense. I'll create a `kind-ensure` (or `kind-recreate`) target that deletes + creates, and keep `kind-create` as-is. The prepare target can call the new one.

acceptance/make/common.mk
oswcab: The wait is a no-op if the node is already Ready (exits immediately). It was added after hitting a case where `cert-manager` pods were scheduled before the node was fully ready, causing scheduling failures. But if you haven't seen that in CI, I'm fine removing it; the `cert-manager` rollout timeout would catch it anyway, just with a less clear error.

oswcab:
> I'm half-way. Leaving some initial feedback. It would be great if we could split this one in multiple PRs.

Done. Please check the other PR here: https://github.com/konflux-ci/namespace-lister/pull/337

acceptance/make/common.mk
filariow: > it was the webhook admission endpoint not accepting connections yet (even after pods reported Ready).

I'd expect the endpoint used by the webhook to be ready if its readiness check are passing in first place 🙃 Maybe bumping to the latest version of cert-manager fix this already. Otherwise, I would expect "setting pods.spec.containers.readinessProbe.successThreshold to an higher value" to give it more time to really get ready.

PR #337, opened 2026-04-24
Fix acceptance test setup for local development

oswcab:
Running acceptance tests locally (`make -C acceptance/test/dumb-proxy prepare`) was failing due to several infrastructure issues:
1. `kube-proxy` crashing with "too many open files" left the cluster in a bad state, and re-running `prepare` failed because the Kind cluster still existed
2. Pods failed to schedule because cert-manager was applied before the node  was fully ready
3. `cert-manager webhook` and `cainjector` crashed in CrashLoopBackOff because the CPU/memory limits were too tight, causing API server connection timeouts
4. With `podman` as the container runtime, `kind load image-archive` via stdin silently failed, and the image was stored with a `localhost/` prefix that didn't match the deployment's image reference


acceptance/make/common.mk
```diff
@@ -50,6 +54,16 @@ deploy-test-infra:
...
+	@for i in $$(seq 1 30); do \
+		if $(KUBECTL) get validatingwebhookconfigurations cert-manager-webhook >/dev/null 2>&1 && \
+		   $(KUBECTL) apply --dry-run=server -f - >/dev/null 2>&1 <<< \
+		   '{"apiVersion":"cert-manager.io/v1","kind":"Issuer","metadata":{"name":"probe","namespace":"default"},"spec":{"selfSigned":{}}}'; then \
+			break; \
+		fi; \
+		echo "  webhook not ready yet, retrying ($$i/30)..."; \
+		sleep 2; \
+	done
 	$(KUBECTL) apply -k $(ROOT_DIR)/dependencies/cluster-issuer/
```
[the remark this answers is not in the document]
oswcab: Done

acceptance/make/common.mk
```diff
@@ -50,6 +54,16 @@ deploy-test-infra:
 		--timeout=300s \
 		-l 'app.kubernetes.io/instance=cert-manager' \
 		-n cert-manager deployment
+	@echo "Waiting for cert-manager webhook to accept connections..."
+	@for i in $$(seq 1 30); do \
+		if $(KUBECTL) get validatingwebhookconfigurations cert-manager-webhook >/dev/null 2>&1 && \
+		   $(KUBECTL) apply --dry-run=server -f - >/dev/null 2>&1 <<< \
+		   '{"apiVersion":"cert-manager.io/v1","kind":"Issuer","metadata":{"name":"probe","namespace":"default"},"spec":{"selfSigned":{}}}'; then \
+			break; \
```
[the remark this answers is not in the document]
oswcab: Done.

acceptance/make/common.mk
```diff
@@ -50,6 +57,22 @@ deploy-test-infra:
...
+			webhook_ready=true; \
+			break; \
+		fi; \
+		echo "  webhook not ready yet, retrying ($$i/30)..."; \
+		sleep 2; \
+	done; \
+	if [ "$$webhook_ready" != "true" ]; then \
+		echo "ERROR: cert-manager webhook did not become ready after 30 attempts"; \
+		exit 1; \
+	fi
```
filariow, line 74: I'm not a fan of adding this complexity. Cert-Manager is a very common dependency. How are others coping with this problem?

Maybe we can rely on readiness/liveness probes on cert-manager's webhook instead?

acceptance/make/common.mk
```diff
@@ -33,10 +33,17 @@ image-build:
 kind-create:
 	$(KIND) create cluster --name "$(KIND_CLUSTER_NAME)" --config kind-config.yaml
 
+.PHONY: kind-recreate
+kind-recreate:
+	-$(KIND) delete cluster --name "$(KIND_CLUSTER_NAME)"
+	$(MAKE) kind-create
+
```
filariow: I'm not sure I like this, but it's kind of a matter of personal preference I guess 😄 

What if instead we add `kind-delete`, leave the `prepare` as it is nowadays, and when we want this behavior we just run `make kind-delete prepare`?

acceptance/make/common.mk
```diff
@@ -33,10 +33,17 @@ image-build:
...
 .PHONY: kind-load-image
-kind-load-image:
-	$(IMAGE_BUILDER) save $(IMG) | \
-		$(KIND) load image-archive --name $(KIND_CLUSTER_NAME) /dev/stdin
+kind-load-image: $(OUT_DIR)
+	$(IMAGE_BUILDER) tag $(IMG) docker.io/library/$(IMG) 2>/dev/null || true
+	$(IMAGE_BUILDER) save docker.io/library/$(IMG) -o $(OUT_DIR)/image.tar
+	$(KIND) load image-archive $(OUT_DIR)/image.tar --name $(KIND_CLUSTER_NAME)
+	rm -f $(OUT_DIR)/image.tar
```
filariow, line 45: podman by default adds the `localhost/` prefix on local builds. We can overcome this problem by setting `IMG ?= localhost/namespace-lister:latest` at the top of this file

filariow (changes requested):

acceptance/make/common.mk
filariow: just noticed we have a similar thread here: https://github.com/konflux-ci/namespace-lister/pull/333/changes#r3195991777

oswcab: Good point. The reason we added this probe is that there's a known race condition: `rollout status` can report the cert-manager webhook deployment as ready before the webhook actually starts accepting API server connections. This causes `kubectl apply -k .../cluster-issuer/` to fail intermittently — the exact flakiness we were trying to fix.

cert-manager's own `cmctl check api --wait=60s` is the official way to handle this, but it adds a tool dependency. The alternative is `kubectl wait --for=condition=Available deployment/cert-manager-webhook -n cert-manager` but that has the same race issue — pod readiness ≠ webhook readiness from the API server's perspective.

That said, I agree the dry-run approach is complex. What do you think about simplifying it to just retry the `kubectl apply -k .../cluster-issuer/` itself on failure? That way we don't need the probe — we just handle the transient failure directly:

```makefile
@for i in $$(seq 1 5); do \
	$(KUBECTL) apply -k $(ROOT_DIR)/dependencies/cluster-issuer/ && break; \
	echo "  cluster-issuer apply failed, retrying ($$i/5)..."; \
	sleep 5; \
done
```

acceptance/make/common.mk
oswcab: Makes sense, done. Added a standalone `kind-delete` target and reverted `prepare` to use `kind-create`. Users who want a fresh start can run `make kind-delete prepare`.

acceptance/make/common.mk
oswcab: Good call, done. Changed `IMG ?= localhost/namespace-lister:latest` and simplified `kind-load-image` by removing the tag workaround.

Note: the new default targets podman. Docker users can override with `IMG=namespace-lister:latest`.

acceptance/make/common.mk
filariow: I like the retry approach more

acceptance/make/common.mk
```diff
@@ -1,5 +1,6 @@
 KIND_CLUSTER_NAME ?= namespace-lister-acceptance-tests
-IMG ?= namespace-lister:latest
+# Default targets podman; docker users can override with IMG=namespace-lister:latest
```
filariow, line 2: I think it should work with docker without having to change `IMG`. Can you give it a try? In that case we can remove the comment. Otherwise, we'll need to handle it a bit better and give better support to docker users

oswcab: I checked this — unfortunately Docker and podman handle `localhost/` differently. Podman auto-prefixes local images with `localhost/`, so `podman build -t namespace-lister:latest` stores it as `localhost/namespace-lister:latest`. But Docker treats `localhost/` as a registry hostname — `docker save localhost/namespace-lister:latest` would fail because Docker doesn't resolve it to the locally built `namespace-lister:latest` (which it stores as `docker.io/library/namespace-lister:latest`).

So the comment is needed to let docker users know they should override IMG. Unless you'd prefer a different approach — e.g., detecting the runtime and setting the prefix conditionally, or keeping `IMG ?= namespace-lister:latest` and handling the podman case differently?

acceptance/make/common.mk
oswcab: Done — replaced the webhook probe with a simple retry loop on the cluster-issuer apply. Much cleaner.

acceptance/make/common.mk
oswcab: Actually, went ahead and implemented runtime detection instead. Now IMG is set conditionally based on IMAGE_BUILDER:

```makefile
IMAGE_BUILDER ?= docker

ifeq ($(IMAGE_BUILDER),podman)
IMG ?= localhost/namespace-lister:latest
else
IMG ?= namespace-lister:latest
endif
```

Works for both runtimes without any manual override needed.

PR #341, opened 2026-04-26
KFLUXINFRA-3536: Add unit tests for Prometheus metrics registration

glevi-rh:
## What

Add unit tests for the Prometheus metric factories and registration logic in metrics_http.go and metrics.go.

- Test all 4 HTTP metric constructors (request latency histogram, request counter, response size histogram, in-flight gauge) for correct name, help text, type, label set, and histogram bucket boundaries
- Test InitRegistry process collector registration
- Test duplicate registration panics for both newHTTPMetrics and InitRegistry

## Why

These files had zero test coverage. The tests verify that metrics are registered with the correct names, labels, and histogram bucket configuration, catching regressions if metric definitions are accidentally changed.

## Validation

- All 13 new tests pass: go test -run TestNamespaceLister . -ginkgo.focus 'HTTP Metrics|InitRegistry'
- Each test uses a fresh prometheus.NewRegistry() -- no global state leaks between tests

### DIFF
--- a/metrics_http_test.go
+++ b/metrics_http_test.go
@@ -1,0 +1,183 @@
+package main
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("HTTP Metrics", func() {
+	var reg *prometheus.Registry
+	var m httpMetrics
+
+	BeforeEach(func() {
+		reg = prometheus.NewRegistry()
+		m = newHTTPMetrics(reg)
+	})
+
+	Describe("registration", func() {
+		It("registers all metrics with correct names", func() {
+			m.requestTiming.WithLabelValues("200", "GET").Observe(0)
+			m.requestCounter.WithLabelValues("200", "GET").Inc()
+			m.responseSize.WithLabelValues("200", "GET").Observe(0)
+			m.inFlightGauge.Set(0)
+
+			families, err := reg.Gather()
+			Expect(err).NotTo(HaveOccurred())
+
+			names := metricFamilyNames(families)
+			Expect(names).To(ConsistOf(
+				"namespace_lister_api_latency",
+				"namespace_lister_api_counter",
+				"namespace_lister_api_response_size",
+				"namespace_lister_api_requests_in_flight",
+			))
+		})
+
+		It("panics on duplicate registration", func() {
+			Expect(func() { newHTTPMetrics(reg) }).To(Panic())
+		})
+	})
+
+	Describe("request latency histogram", func() {
+		var mf *dto.MetricFamily
+
+		BeforeEach(func() {
+			m.requestTiming.WithLabelValues("200", "GET").Observe(0.001)
+			mf = findFamily(reg, "namespace_lister_api_latency")
+		})
+
+		It("has correct help text and type", func() {
+			Expect(mf.GetHelp()).To(Equal("Latency of requests"))
+			Expect(mf.GetType()).To(Equal(dto.MetricType_HISTOGRAM))
+		})
+
+		It("has expected bucket boundaries", func() {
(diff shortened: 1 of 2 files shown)


metrics_http_test.go
```diff
@@ -0,0 +1,178 @@
...
+		Expect(err).NotTo(HaveOccurred())
+
+		names := metricFamilyNames(families)
+		Expect(names).To(ContainElements(
+			"namespace_lister_api_latency",
+			"namespace_lister_api_counter",
+			"namespace_lister_api_response_size",
+			"namespace_lister_api_requests_in_flight",
+		))
+	})
```
meyrevived: It() blocks don't usually live outside of Describe() blocks. It messes up Ginkgo a little bit.

metrics_http_test.go
```diff
@@ -0,0 +1,178 @@
...
+		m.requestTiming.WithLabelValues("200", "GET").Observe(0)
+		m.requestCounter.WithLabelValues("200", "GET").Inc()
+		m.responseSize.WithLabelValues("200", "GET").Observe(0)
+		m.inFlightGauge.Set(0)
+
+		families, err := reg.Gather()
+		Expect(err).NotTo(HaveOccurred())
+
+		names := metricFamilyNames(families)
+		Expect(names).To(ContainElements(
```
meyrevived: Do we expect _some_ of these strings or _all_ of them? 
If only some of these strings are what's expected, the way this assertion is written is fine, but if we need `names` to have all of them, [gomega documentation](https://onsi.github.io/gomega/#containelementelement-any) has an explanation on how to do it, look for the paragraph that starts with `"In addition, there are occasions when you need to grab (all) matching contained elements"`

metrics_http_test.go
```diff
@@ -0,0 +1,178 @@
...
+			m := newHTTPMetrics(reg)
+			m.requestTiming.WithLabelValues("200", "GET").Observe(0)
+
+			mf := findFamily(reg, "namespace_lister_api_latency")
+			Expect(mf.GetHelp()).To(Equal("Latency of requests"))
+			Expect(mf.GetType()).To(Equal(dto.MetricType_HISTOGRAM))
+		})
+
+		It("has expected bucket boundaries", func() {
+			m := newHTTPMetrics(reg)
```
meyrevived: If it's a repeating variable that's always setting up things for the unit tests to run, it's better to move them to a BeforeEach() right under  `var reg *prometheus.Registry`

metrics_http_test.go
```diff
@@ -0,0 +1,178 @@
...
+			boundaries := bucketBoundaries(buckets)
+			Expect(boundaries[0]).To(BeNumerically("~", 1e-9, 1e-15))
+			Expect(boundaries[len(boundaries)-1]).To(BeNumerically("==", 60.0))
+		})
+
+		It("uses code and method labels", func() {
+			m := newHTTPMetrics(reg)
+			m.requestTiming.WithLabelValues("200", "GET").Observe(0.1)
+
+			mf := findFamily(reg, "namespace_lister_api_latency")
```
meyrevived: Same comment about repeating code as the one about line 48

meyrevived (changes requested):
Good job - it's a but short, but tha'ts expected from a tiny reo.
Some minor corrections requirew

meyrevived (approved):
/lgtm

metrics_test.go
```diff
@@ -0,0 +1,27 @@
...
+		Expect(err).NotTo(HaveOccurred())
+
+		names := metricFamilyNames(families)
+		Expect(names).To(ContainElement(HavePrefix("namespace_lister_process_")))
+	})
+
+	It("panics on duplicate registration", func() {
+		reg := prometheus.NewRegistry()
+		InitRegistry(reg)
+		Expect(func() { InitRegistry(reg) }).To(Panic())
```
filariow, line 25: Is this right? it looks like a duplicate

glevi-rh: What do you mean by duplicate?

filariow: InitRegistry is called twice. I think L24 is a leftover.

filariow: oh now I get it. It's expected. ok! 😄

PR #359, opened 2026-05-07
feat: add cache acceptance test suite

oswcab:
Add BDD tests for namespace cache behavior:
- Namespace created after startup becomes visible
- Namespace deleted while service is running becomes invisible
- RoleBinding added grants access to existing namespace
- RoleBinding removed revokes access
- Bulk concurrent namespace and RBAC changes

Contributes to: KFLUXINFRA-3243

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -64,7 +64,7 @@ jobs:
 
       - uses: kubernetes-sigs/kwok@main
         with:
-          command: 'kwokctl'
+          command: "kwokctl"
 
       - name: Set up kwokctl
         uses: kubernetes-sigs/kwok@main
@@ -111,7 +111,7 @@ jobs:
           echo "##[group] Preparing environment"
             make -C "acceptance/test/${{ matrix.proxy }}" prepare
           echo "##[endgroup]"
-          echo "##[group] Running tests"
+          echo "##[group] Running tests (${{ matrix.proxy }})"
             make -C "acceptance/test/${{ matrix.proxy }}" test
           echo "##[endgroup]"
 
@@ -143,7 +143,6 @@ jobs:
           flags: e2e-tests
           directory: coverage-output
           fail_ci_if_error: false
-
 
   go-tidy:
     name: Tidy go mod
--- a/acceptance/pkg/suite/steps_cache.go
+++ b/acceptance/pkg/suite/steps_cache.go
@@ -1,0 +1,234 @@
+package suite
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"slices"
+	"time"
+
+	"github.com/cucumber/godog"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	tcontext "github.com/konflux-ci/namespace-lister/acceptance/pkg/context"
+	arest "github.com/konflux-ci/namespace-lister/acceptance/pkg/rest"
+)
+
+func InjectCacheSteps(ctx *godog.ScenarioContext) {
+	ctx.Given(`^a namespace "([^"]*)" exists without access for the current user$`, aNamespaceExistsWithoutAccess)
+	ctx.When(`^a new namespace "([^"]*)" is created with access for the current user$`, aNewNamespaceIsCreatedWithAccess)
+	ctx.When(`^the namespace "([^"]*)" is deleted$`, theNamespaceIsDeleted)
+	ctx.When(`^a RoleBinding granting access is added in namespace "([^"]*)"$`, aRoleBindingIsAddedInNamespace)
+	ctx.When(`^the RoleBinding is removed from namespace "([^"]*)"$`, theRoleBindingIsRemovedFromNamespace)
+	ctx.When(`^"(\d+)" namespaces with access are created and "(\d+)" existing namespaces are deleted$`, bulkNamespaceChanges)
+
+	ctx.Then(`^the user can see namespace "([^"]*)" in the list$`, theUserCanSeeNamespaceInTheList)
+	ctx.Then(`^the user cannot see namespace "([^"]*)" in the list$`, theUserCannotSeeNamespaceInTheList)
+}
+
+func aNamespaceExistsWithoutAccess(ctx context.Context, nsName string) (context.Context, error) {
+	run := tcontext.RunId(ctx)
+	user := tcontext.User(ctx)
+	fullName := fmt.Sprintf("run-%s-%s", run, nsName)
+
+	cli, err := arest.BuildDefaultHostClient()
+	if err != nil {
+		return ctx, err
+	}
+
+	ns := newTestNamespace(fullName, run)
+	if err := cli.Create(ctx, ns); err != nil && !errors.IsAlreadyExists(err) {
+		return ctx, err
+	}
+
+	// ensure no leftover RoleBinding from a previous run grants access
+	rb := newAccessRoleBinding(fmt.Sprintf("run-%s-%s-access", run, nsName), fullName, user.AsSubject())
+	if err := cli.Delete(ctx, rb); err != nil && !errors.IsNotFound(err) {
+		return ctx, err
+	}
+
+	return ctx, nil
+}
+
+func aNewNamespaceIsCreatedWithAccess(ctx context.Context, nsName string) (context.Context, error) {
+	run := tcontext.RunId(ctx)
+	user := tcontext.User(ctx)
+	fullName := fmt.Sprintf("run-%s-%s", run, nsName)
+
+	cli, err := arest.BuildDefaultHostClient()
+	if err != nil {
+		return ctx, err
+	}
+
+	ns := newTestNamespace(fullName, run)
+	if err := cli.Create(ctx, ns); err != nil && !errors.IsAlreadyExists(err) {
+		return ctx, err
+	}
+
+	rb := newAccessRoleBinding(fmt.Sprintf("run-%s-%s-access", run, nsName), fullName, user.AsSubject())
+	if err := cli.Create(ctx, rb); err != nil && !errors.IsAlreadyExists(err) {
+		return ctx, err
+	}
+
+	nn := tcontext.Namespaces(ctx)
+	nn = append(nn, *ns)
+	return tcontext.WithNamespaces(ctx, nn), nil
+}
+
+func theNamespaceIsDeleted(ctx context.Context, nsName string) (context.Context, error) {
+	run := tcontext.RunId(ctx)
+	fullName := fmt.Sprintf("run-%s-%s", run, nsName)
+
+	cli, err := arest.BuildDefaultHostClient()
+	if err != nil {
+		return ctx, err
+	}
+
+	ns := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: fullName}}
+	if err := cli.Delete(ctx, ns); err != nil && !errors.IsNotFound(err) {
+		return ctx, err
+	}
+
+	nn := tcontext.Namespaces(ctx)
+	nn = slices.DeleteFunc(nn, func(n corev1.Namespace) bool {
+		return n.Name == fullName
+	})
+	return tcontext.WithNamespaces(ctx, nn), nil
+}
+
+func aRoleBindingIsAddedInNamespace(ctx context.Context, nsName string) (context.Context, error) {
+	run := tcontext.RunId(ctx)
+	user := tcontext.User(ctx)
+	fullName := fmt.Sprintf("run-%s-%s", run, nsName)
+
+	cli, err := arest.BuildDefaultHostClient()
+	if err != nil {
+		return ctx, err
+	}
+
+	rb := newAccessRoleBinding(fmt.Sprintf("run-%s-%s-access", run, nsName), fullName, user.AsSubject())
+	if err := cli.Create(ctx, rb); err != nil && !errors.IsAlreadyExists(err) {
+		return ctx, err
+	}
+
+	// track the namespace as expected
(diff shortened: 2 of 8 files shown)


.github/workflows/ci.yaml
```diff
@@ -107,10 +107,18 @@ jobs:
...
             make -C "acceptance/test/${{ matrix.proxy }}" test
           echo "##[endgroup]"
 
+      - name: Upload test results
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: acceptance-test-results-${{ matrix.proxy }}
+          path: acceptance/test/${{ matrix.proxy }}/out/
+          retention-days: 7
```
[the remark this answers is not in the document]
oswcab: Done

acceptance/pkg/suite/steps_cache.go
```diff
@@ -0,0 +1,330 @@
...
+		}
+
+		nn := tcontext.Namespaces(ctx)
+		if len(ann.Items) != expected {
+			names := make([]string, len(ann.Items))
+			for i, n := range ann.Items {
+				names[i] = n.Name
+			}
+			log.Printf("expected %d namespaces, got %d: %v", expected, len(ann.Items), names)
+			return false, nil
```
meyrevived: Qodo also flagged this under flaky tests - basically, `theUserSeesExactlyNNamespaces` compared `len(ann.Items) `(all namespaces visible to the user) against expected, without scoping to the current test run. Leftover namespaces from previous failed runs would inflate the count and cause flaky failures.
Suggesting a fix of filtering `ann.Items` by the `namespace-lister/test-run` label (which matches the current scenario's run ID) before comparing counts. The filtered slice `runItems` replaces `ann.Items` in both the count check and the tracked-namespace membership check. The log message now includes the run ID for easier debugging.

```suggestion
func theUserSeesExactlyNNamespaces(ctx context.Context, expected int) (context.Context, error) {
	run := tcontext.RunId(ctx)
	cli, err := tcontext.InvokeBuildUserClientFunc(ctx)
	if err != nil {
		return ctx, err
	}

	return ctx, wait.PollUntilContextTimeout(ctx, 2*time.Second, 2*time.Minute, true, func(ctx context.Context) (done bool, err error) {
		ann := corev1.NamespaceList{}
		if err := cli.List(ctx, &ann); err != nil {
			log.Printf("error listing namespaces: %v", err)
			return false, nil
		}
		
		// Scope to the current test run to avoid counting leftover
		// namespaces from previous runs in shared environments.
		var runItems []corev1.Namespace
		for _, n := range ann.Items {
			if n.Labels["namespace-lister/test-run"] == run {
				runItems = append(runItems, n)
			}
		}

		nn := tcontext.Namespaces(ctx)
		if len(ann.Items) != expected {
			names := make([]string, len(ann.Items))
			for i, n := range ann.Items {
				names[i] = n.Name
			}
			log.Printf("expected %d namespaces for run %s, got %d: %v", expected, len(ann.Items), names)
			return false, nil
```

acceptance/pkg/suite/steps_cache.go
```diff
@@ -0,0 +1,330 @@
...
+				Namespace: name,
+			},
+			RoleRef: rbacv1.RoleRef{
+				Kind:     "ClusterRole",
+				Name:     "namespace-get",
+				APIGroup: rbacv1.GroupName,
+			},
+			Subjects: []rbacv1.Subject{user.AsSubject()},
+		}
+		if err := cli.Create(ctx, rb); err != nil {
```
meyrevived: The bulk namespace and RoleBinding creates fail hard on any error (`return ctx, err`) without handling `AlreadyExists`. Every other `create` in the file uses `!errors.IsAlreadyExists(err)` to tolerate pre-existing resources.

```suggestion
		if err := cli.Create(ctx, ns); err != nil && !errors.IsAlreadyExists(err) {
			return ctx, err
		}

		rb := &rbacv1.RoleBinding{
			ObjectMeta: metav1.ObjectMeta{
				Name:      fmt.Sprintf("run-%s-bulk-%d-%d-access", run, tn, i),
				Namespace: name,
			},
			RoleRef: rbacv1.RoleRef{
				Kind:     "ClusterRole",
				Name:     "namespace-get",
				APIGroup: rbacv1.GroupName,
			},
			Subjects: []rbacv1.Subject{user.AsSubject()},
		}
		if err := cli.Create(ctx, rb); err != nil && !errors.IsAlreadyExists(err) {
```

acceptance/pkg/suite/steps_cache.go
```diff
@@ -0,0 +1,330 @@
...
+		if deleted >= toDelete {
+			break
+		}
+		ns := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: nn[i].Name}}
+		if err := cli.Delete(ctx, ns); err != nil && !errors.IsNotFound(err) {
+			return ctx, err
+		}
+		deleted++
+	}
+	nn = nn[deleted:]
```
meyrevived: If `toDelete` exceeds `len(nn)`, the function deletes fewer namespaces than requested without reporting the discrepancy. The loop just stops when it runs out of tracked namespace. The current test scenario calls with `toDelete=5` on a set of exactly 5 namespaces, so it works. 
But the function's signature accepts _any_ `toDelete` value, making the silent truncation a latent bug. A future scenario author might reasonably expect `5 namespaces with access are created and 3 existing namespaces are deleted` to leave 7, but if tracked state drifts, the deletion count will silently differ.

```suggestion
	nn := tcontext.Namespaces(ctx)
	if toDelete > len(nn) {
		return ctx, fmt.Errorf("requested deletion of %d namespaces but only %d are tracked", toDelete, len(nn))
	}
	for i := range toDelete {
		ns := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: nn[i].Name}}
		if err := cli.Delete(ctx, ns); err != nil && !errors.IsNotFound(err) {
			return ctx, err
		}
	}
	nn = nn[toDelete:]
```

acceptance/pkg/suite/steps_cache.go
```diff
@@ -0,0 +1,330 @@
...
+	ns := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: fullName,
+			Labels: map[string]string{
+				"konflux.ci/type":           "user",
+				"namespace-lister/scope":    "acceptance-tests",
+				"namespace-lister/test-run": run,
+			},
+		},
+	}
```
meyrevived: This NameSpace is created in the exact same way with the exact same values 3 times in this file. This should be a var or a const and appear only once in the file.

acceptance/pkg/suite/steps_cache.go
meyrevived, line 1: Both this file and `acceptance/pkg/suite/steps.go` repeatedly construct namespaces and RoleBindings with the same structure. 
I suggest creating a `steps_and_cache_helper.go` with helpers who do that so we can reduce duplication across both files and make the test suite more maintainable.

meyrevived (changes requested):
I am, again, very sorry I didn't see this one before.
I left a few code change suggestions for issues I found. 
This is mostly a testing infrastructure PR, not much coverage gaps to suggest without testing our testing-mocks.

acceptance/pkg/suite/steps_cache.go
oswcab: Done.

acceptance/pkg/suite/steps_cache.go
oswcab: Done.

acceptance/pkg/suite/steps_cache.go
oswcab: Done.

acceptance/pkg/suite/steps_cache.go
oswcab: Done.

acceptance/pkg/suite/steps_cache.go
oswcab: Done.

Makefile
```diff
@@ -70,6 +70,13 @@ test-perf: ## Run performance tests
 		--keep-going --procs=1 --flake-attempts 2 --output-dir=$(PERF_OUT_DIR)
 	# -$(PERF_CLUSTER_PROVIDER) delete cluster --name namespace-lister-perf-test
 
+.PHONY: test-e2e
+test-e2e: ## Run all acceptance test suites end-to-end.
+	$(MAKE) -C acceptance/test/dumb-proxy prepare
+	$(MAKE) -C acceptance/test/dumb-proxy test
+	$(MAKE) -C acceptance/test/smart-proxy prepare
+	$(MAKE) -C acceptance/test/smart-proxy test
+
```
filariow: what do we need this for?

acceptance/test/dumb-proxy/features/cache.feature
```diff
@@ -0,0 +1,34 @@
+@cache @serial
```
filariow, line 1: what do we need these annotation for?

acceptance/test/smart-proxy/features/cache.feature
```diff
@@ -0,0 +1,34 @@
...
+    When a new namespace "to-revoke" is created with access for the current user
+    Then the user can see namespace "to-revoke" in the list
+    When the RoleBinding is removed from namespace "to-revoke"
+    Then the user cannot see namespace "to-revoke" in the list
+
+  Scenario: Bulk concurrent namespace and RBAC changes
+    Given User has access to "5" namespaces
+    Then the User can retrieve only the namespaces they have access to
+    When 5 namespaces with access are created and 5 existing namespaces are deleted
+    Then the user sees exactly 5 namespaces
```
filariow: these are not respecting the convention that dynamic values are quoted


```suggestion
    When "5" namespaces with access are created and "5" existing namespaces are deleted
    Then the user sees exactly "5" namespaces
```

acceptance/test/smart-proxy/features/cache.feature
```diff
@@ -0,0 +1,34 @@
...
+    Given User has access to "3" namespaces
+    When a new namespace "to-revoke" is created with access for the current user
+    Then the user can see namespace "to-revoke" in the list
+    When the RoleBinding is removed from namespace "to-revoke"
+    Then the user cannot see namespace "to-revoke" in the list
+
+  Scenario: Bulk concurrent namespace and RBAC changes
+    Given User has access to "5" namespaces
+    Then the User can retrieve only the namespaces they have access to
+    When 5 namespaces with access are created and 5 existing namespaces are deleted
```
filariow: let's use a different higher number, so we are sure something changed

```suggestion
    When 10 namespaces with access are created and 5 existing namespaces are deleted
```

acceptance/test/smart-proxy/features/cache.feature
filariow: it would be good to check that the ones that the user can see now are the ones recently created

acceptance/test/smart-proxy/features/cache.feature
```diff
@@ -0,0 +1,34 @@
+@cache @serial
```
filariow, line 1: ditto

acceptance/pkg/suite/steps_cache.go
```diff
@@ -0,0 +1,274 @@
...
+	for i := range toDelete {
+		ns := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: nn[i].Name}}
+		if err := cli.Delete(ctx, ns); err != nil && !errors.IsNotFound(err) {
+			return ctx, err
+		}
+	}
+	nn = nn[toDelete:]
+
+	// create new namespaces
+	tn := time.Now().Unix()
```
filariow: I don't get why we need this. I think we can remove it 😄

filariow (changes requested):

Makefile
oswcab: It was meant as a convenience target to run both acceptance suites locally in one command. Removing it since the CI workflow already handles this.

acceptance/test/dumb-proxy/features/cache.feature
oswcab: `@cache` allows selective test execution (e.g. `--godog.tags=cache`). `@serial` forces sequential execution since cache scenarios mutate shared state (creating/deleting namespaces and RoleBindings) and would interfere with each other if run concurrently.

acceptance/pkg/suite/steps_cache.go
oswcab: You're right, the run ID (scenario ID) already provides uniqueness. The timestamp was redundant. Removed.

acceptance/test/smart-proxy/features/cache.feature
oswcab: Done. Quoted the numbers and changed to create 7 / delete 5 so the final count (7) proves a change happened. Also replaced the count-only assertion with `the User/ServiceAccount can retrieve only the namespaces they have access to` which verifies both count and membership — ensuring the visible namespaces are exactly the recently created ones.

acceptance/pkg/suite/steps_cache.go
```diff
@@ -0,0 +1,274 @@
...
+	if err != nil {
+		return ctx, err
+	}
+
+	ns := newTestNamespace(fullName, run)
+	if err := cli.Create(ctx, ns); err != nil && !errors.IsAlreadyExists(err) {
+		return ctx, err
+	}
+
+	return ctx, nil
```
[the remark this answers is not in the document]
oswcab: Done.

acceptance/test/smart-proxy/features/cache.feature
oswcab: Done

acceptance/test/smart-proxy/features/cache.feature
oswcab: Done.

acceptance/test/smart-proxy/features/cache.feature
oswcab: Please see above.

meyrevived (approved):
/lgtm

oswcab:
Merged as part of PR #384

PR #363, opened 2026-05-07
Add unit tests for resource cache transforms and access cache startup

glevi-rh:
## Summary
- Add unit tests for transform functions (`TrimAnnotations`, `MergeTransformFunc`, `TrimRole`, `TrimClusterRole`) in `internal/resourcecache/internal/transform/`
- Move access cache helpers (`BuildAndRegisterAccessCacheMetrics`, `GetResyncPeriodFromEnvOrZero`) from root to `internal/resourcecache/` and add unit tests
- Remove `export_test.go` — all tested functions are now exported in their respective packages

## Test plan
- [x] 10 transform tests pass
- [x] 6 access cache tests pass
- [x] Existing tests unaffected
- [x] No `export_test.go`

### DIFF
--- a/internal/resourcecache/access_cache_test.go
+++ b/internal/resourcecache/access_cache_test.go
@@ -1,0 +1,92 @@
+package resourcecache_test
+
+import (
+	"context"
+	"os"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/konflux-ci/namespace-lister/internal/constants"
+	"github.com/konflux-ci/namespace-lister/internal/resourcecache"
+)
+
+var _ = Describe("Access Cache", func() {
+	Describe("BuildAndRegisterAccessCacheMetrics", func() {
+		var reg *prometheus.Registry
+
+		BeforeEach(func() {
+			reg = prometheus.NewRegistry()
+		})
+
+		It("returns nil metrics when registry is nil", func() {
+			// when
+			metrics, err := resourcecache.BuildAndRegisterAccessCacheMetrics(nil)
+
+			// then
+			Expect(err).NotTo(HaveOccurred())
+			Expect(metrics).To(BeNil())
+		})
+
+		It("registers and returns metrics with a valid registry", func() {
+			// when
+			metrics, err := resourcecache.BuildAndRegisterAccessCacheMetrics(reg)
+
+			// then
+			Expect(err).NotTo(HaveOccurred())
+			Expect(metrics).NotTo(BeNil())
+		})
+
+		It("returns an error on duplicate registration", func() {
+			// given
+			_, err := resourcecache.BuildAndRegisterAccessCacheMetrics(reg)
+			Expect(err).NotTo(HaveOccurred())
+
+			// when
+			_, err = resourcecache.BuildAndRegisterAccessCacheMetrics(reg)
+
+			// then
+			Expect(err).To(BeAssignableToTypeOf(prometheus.AlreadyRegisteredError{}))
+		})
+	})
+
+	Describe("GetResyncPeriodFromEnvOrZero", func() {
+		const envKey = constants.EnvCacheResyncPeriod
+
+		It("returns zero when env is not set", func(ctx context.Context) {
+			// given
+			GinkgoT().Setenv(envKey, "dummy")
+			Expect(os.Unsetenv(envKey)).To(Succeed()) //nolint:usetesting // need truly-unset env for this test
+
+			// when
+			d := resourcecache.GetResyncPeriodFromEnvOrZero(ctx)
+
+			// then
+			Expect(d).To(Equal(time.Duration(0)))
+		})
+
+		It("parses a valid duration string", func(ctx context.Context) {
+			// given
+			GinkgoT().Setenv(envKey, "5m")
+
+			// when
+			d := resourcecache.GetResyncPeriodFromEnvOrZero(ctx)
+
+			// then
+			Expect(d).To(Equal(5 * time.Minute))
+		})
+
+		It("returns zero for an invalid duration string", func(ctx context.Context) {
+			// given
+			GinkgoT().Setenv(envKey, "not-a-duration")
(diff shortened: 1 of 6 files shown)


internal/resourcecache/access_cache_test.go
```diff
@@ -0,0 +1,87 @@
...
+	Describe("getResyncPeriodFromEnvOrZero", func() {
+		const envKey = constants.EnvCacheResyncPeriod
+
+		It("returns zero when env is not set", func(ctx context.Context) {
+			// when
+			d := namespacelister.GetResyncPeriodFromEnvOrZero(ctx)
+
+			// then
+			Expect(d).To(Equal(time.Duration(0)))
+		})
```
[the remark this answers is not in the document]
glevi-rh: I changed the logic there to address this issue.

internal/resourcecache/internal/transform/transformfuncs_test.go
```diff
@@ -0,0 +1,211 @@
...
+			// given
+			ns := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "test-ns"}}
+
+			// when
+			result, err := transform.TrimAnnotations()(ns)
+
+			// then
+			Expect(err).NotTo(HaveOccurred())
+			Expect(result.(*corev1.Namespace).Annotations).To(BeNil())
+		})
```
filariow, line 49: we can use a DescribeTable here to simplify this code

internal/resourcecache/internal/transform/transformfuncs_test.go
```diff
@@ -0,0 +1,211 @@
...
+	})
+
+	Describe("MergeTransformFunc", func() {
+		It("chains multiple transforms in order", func() {
+			// given
+			addLabel := func(i interface{}) (interface{}, error) {
```
filariow, line 55: in modern go `any` is preferred to `interface{}`

internal/resourcecache/internal/transform/transformfuncs_test.go
```diff
@@ -0,0 +1,211 @@
...
+
+			// when
+			result, err := transform.TrimAnnotations()(ns)
+
+			// then
+			Expect(err).NotTo(HaveOccurred())
+			out := result.(*corev1.Namespace)
+			Expect(out.Annotations).To(BeNil())
+			Expect(out.Labels).To(Equal(map[string]string{"team": "infra"}))
+			Expect(out.Name).To(Equal("test-ns"))
```
filariow, line 36: I feel we can improve this. Something like what follows:


```suggestion
			// given
			ns := &corev1.Namespace{
				ObjectMeta: metav1.ObjectMeta{
					Name:        "test-ns",
					Labels:      map[string]string{"team": "infra"},
					Annotations: map[string]string{"note": "value"},
				},
			}

			// when
			result, err := transform.TrimAnnotations()(ns.DeepCopy())

			// then - no error produced
			Expect(err).NotTo(HaveOccurred())
			out := result.(*corev1.Namespace)
			Expect(out).NotTo(BeNil())

			// then - the only difference is the absence of the annotation
			Expect(ns.DeepCopy()).To(
				WithTransform(func(ns *corev1.Namespace) *corev1.Namespace {
					ns.Annotations = nil
					return ns
				}, BeEquivalentTo(out)))
```

internal/resourcecache/internal/transform/transformfuncs_test.go
```diff
@@ -0,0 +1,211 @@
...
+
+			// when
+			result, err := merged(ns)
+
+			// then
+			Expect(err).NotTo(HaveOccurred())
+			out := result.(*corev1.Namespace)
+			Expect(out.Labels).To(HaveKeyWithValue("added", "true"))
+			Expect(out.Spec.Finalizers).To(BeEmpty())
+		})
```
filariow, line 82: the `MergeTransformFunc` has access to the `Namespace`. I don't think this is the appropriate way to test its behavior. If it (by mistake) starts doing some changes to the fields used in this test to validate it's working, we won't catch it. It's really unlikely to happen as per a human mistake, still this makes me think we can improve the test case:

```suggestion
		var ctrl *gomock.Controller

		BeforeEach(func() {
			ctrl = gomock.NewController(GinkgoT())
			defer ctrl.Finish()
		})

		It("chains multiple transforms in order", func() {
			// given
			ns := &corev1.Namespace{
				ObjectMeta: metav1.ObjectMeta{Name: "test"},
				Spec:       corev1.NamespaceSpec{Finalizers: []corev1.FinalizerName{corev1.FinalizerKubernetes}},
			}
			tf1 := mocks.NewMockTransformFunc(ctrl)
			tf2 := mocks.NewMockTransformFunc(ctrl)
			gomock.InOrder(
				tf1.EXPECT().TransformFunc(ns).Return(ns, nil).Times(1),
				tf2.EXPECT().TransformFunc(ns).Return(ns, nil).Times(1),
			)

			// when
			merged := transform.MergeTransformFunc(tf1.TransformFunc, tf2.TransformFunc)
			Expect(merged).NotTo(BeNil())
			result, err := merged(ns)

			// then
			Expect(err).NotTo(HaveOccurred())
			out := result.(*corev1.Namespace)
			Expect(out).NotTo(BeNil())
			Expect(out).To(BeEquivalentTo(ns))
		})
```

you also need to add a in `interface_test.go` file with the following content and run `make generate-code`

```golang
package transform_test

//go:generate mockgen -source=interfaces_test.go -destination=mocks/transformfunc_interface.go -package=mocks

type TransformFunc interface {
	TransformFunc(interface{}) (interface{}, error) // "k8s.io/client-go/tools/cache".TransformFunc
}
```

filariow (changes requested):
Thanks for working on this! 😄 

I reviewed the `transform` pkg only. Please split this PR in two: it will make the review easier.

I proposed some changes that applies to the other tests too. If you find them appropriate, please use the same approach for the other tests too.

glevi-rh:
The PR was split to #371 and #372

glevi-rh:
The PR was split to #371 and #372

glevi-rh:
Closing in favor of split to #371 and #372

PR #364, opened 2026-05-07
KFLUXINFRA-3510: add unit tests for addMetricsMiddleware

glevi-rh:
## Summary
- Move `metrics_http.go` to `internal/http/middleware/metrics_http.go`
- Add unit tests for `addMetricsMiddleware` covering nil registerer passthrough, metrics instrumentation, label recording, and in-flight request tracking
- Tests verify all four metric families (latency, counter, response size, in-flight gauge) are registered and populated after serving HTTP requests

## Test plan
- [x] All new tests pass (`go test --ginkgo.focus="addMetricsMiddleware"`)
- [x] Full test suite passes (excluding pre-existing cluster-dependent performance test)


internal/http/middleware/metrics_http_test.go
```diff
@@ -145,6 +148,99 @@ var _ = Describe("HTTP Metrics", func() {
...
+
+		families, err := reg.Gather()
+		Expect(err).NotTo(HaveOccurred())
+
+		names := metricFamilyNames(families)
+		Expect(names).To(ContainElement("namespace_lister_api_latency"))
+		Expect(names).To(ContainElement("namespace_lister_api_counter"))
+		Expect(names).To(ContainElement("namespace_lister_api_response_size"))
+		Expect(names).To(ContainElement("namespace_lister_api_requests_in_flight"))
+	})
```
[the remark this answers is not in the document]
glevi-rh: We intentionally use a real registry — it validates metric names, labels, and buckets end-to-end. A mock would only assert that Register was called, missing actual bugs.

glevi-rh:
Qodo.ai relevant suggestions were implemented.

glevi-rh:
> ## [Codecov](https://app.codecov.io/gh/konflux-ci/namespace-lister/pull/364?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=konflux-ci) Report
> 
> ❌ Patch coverage is `50.00000%` with `1 line` in your changes missing coverage. Please review. ✅ Project coverage is 56.51%. Comparing base ([`bb70d9e`](https://app.codecov.io/gh/konflux-ci/namespace-lister/commit/bb70d9ec086170b825194574208c579263452743?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=konflux-ci)) to head ([`f937105`](https://app.codecov.io/gh/konflux-ci/namespace-lister/commit/f937105eb593dfd2f274dc79233e46ddde54415e?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=konflux-ci)). ⚠️ Report is 3 commits behind head on main.
> [Files with missing lines](https://app.codecov.io/gh/konflux-ci/namespace-lister/pull/364?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=konflux-ci) 	Patch % 	Lines
> [http_api_server.go](https://app.codecov.io/gh/konflux-ci/namespace-lister/pull/364?src=pr&el=tree&filepath=http_api_server.go&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=konflux-ci#diff-aHR0cF9hcGlfc2VydmVyLmdv) 	0.00% 	[1 Missing ⚠️ ](https://app.codecov.io/gh/konflux-ci/namespace-lister/pull/364?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=konflux-ci)
> Additional details and impacted files
> 
> 		
> 		
> 
> 		
> 		
> 		
> 
> 🚀 New features to boost your workflow:

Testing it means we need to create a full HTTP server with a mock authenticator, mock lister, and making real HTTP requests

metrics_http_test.go
meyrevived, line 1: Lots of code in this file didn't move to the file under the new loction in `internal/http/middleware`, was that intentional?

internal/http/middleware/metrics_http_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+var _ = Describe("AddMetricsMiddleware", func() {
+	dummyHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+		_, _ = w.Write([]byte("ok"))
+	})
+
+	It("returns the original handler when registerer is nil", func() {
+		h := middleware.AddMetricsMiddleware(nil, dummyHandler)
+		Expect(fmt.Sprintf("%p", h)).To(Equal(fmt.Sprintf("%p", dummyHandler)))
```
meyrevived, line 25: Technically just a nice to have, but if there's something already out there to check strings, why not

```suggestion
		Expect(h).Should(BeIdenticalTo(dummyHandler))
```

internal/http/middleware/middleware_suite_test.go
```diff
@@ -0,0 +1,47 @@
...
+	RunSpecs(t, "Middleware Suite")
+}
+
+func metricFamilyNames(families []*dto.MetricFamily) []string {
+	names := make([]string, len(families))
+	for i, f := range families {
+		names[i] = f.GetName()
+	}
+	return names
+}
```
meyrevived: This code already exists in _metrics_http_test_, doesn't it?
If it's common code, you can have a new file with this test, under a `testing/utils.go` file where both of these files can approach it and the code will be more DRY.

meyrevived (changes requested):
Nice work, Gal, I have a few notes though

metrics_http_test.go
glevi-rh: No it wasn't, good catch, thanks for that!

internal/http/middleware/metrics_http_test.go
glevi-rh: We can't use `BeIdenticalTo` here, it resolves to `==` on a non-nil function value.

internal/http/middleware/middleware_suite_test.go
glevi-rh: You are right to point that out, but this is the only case of function duplication and it used only in 2 different places. In my opinion adding a dedicated file and package for that function only introduces more complexity, what do you think?

meyrevived: If we need to change anything in this function, even the smallest thing, we'll have to change it in two places. That's more complex and tedious.
Also, e2e tests will need to be written for this repo and their code will sit under `testing/e2e/ `so there will be a change in the repo's structure and that will be dedicated to testing.

internal/http/middleware/metrics_http_test.go
```diff
@@ -0,0 +1,204 @@
...
+		})
+
+		It("uses code and method labels", func() {
+			labels := labelMap(mf.GetMetric()[0])
+			Expect(labels).To(HaveKeyWithValue("code", "200"))
+			Expect(labels).To(HaveKeyWithValue("method", "get"))
+		})
+	})
+
+	Describe("in-flight gauge", func() {
```
meyrevived, line 180: Added a little scenario not covered by the current tests

```suggestion
	})
	
	It("records the response body size", func() {
			reg := prometheus.NewRegistry()
			knownBody := "hello world"
			sizedHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
				w.WriteHeader(http.StatusOK)
				_, _ = w.Write([]byte(knownBody))
			})

			h := middleware.AddMetricsMiddleware(reg, sizedHandler)

			rec := httptest.NewRecorder()
			req := httptest.NewRequest(http.MethodGet, "/", nil)
			h.ServeHTTP(rec, req)

			mf := findFamily(reg, "namespace_lister_api_response_size")
			hist := mf.GetMetric()[0].GetHistogram()
			Expect(*hist.SampleCount).To(BeEquivalentTo(1))
			Expect(*hist.SampleSum).To(BeEquivalentTo(float64(len(knownBody))))
		})
	})

	Describe("in-flight gauge", func() {
```

meyrevived (changes requested):
I added another coverage bump oppertunity

internal/http/middleware/metrics_http_test.go
glevi-rh: Applied in the last commit

internal/http/middleware/metrics_http_test.go
```diff
@@ -0,0 +1,224 @@
...
+		h := middleware.AddMetricsMiddleware(reg, sizedHandler)
+
+		rec := httptest.NewRecorder()
+		req := httptest.NewRequest(http.MethodGet, "/", nil)
+		h.ServeHTTP(rec, req)
+
+		mf := findFamily(reg, "namespace_lister_api_response_size")
+		hist := mf.GetMetric()[0].GetHistogram()
+		Expect(*hist.SampleCount).To(BeEquivalentTo(1))
+		Expect(*hist.SampleSum).To(BeEquivalentTo(float64(len(knownBody))))
```
meyrevived:
```suggestion
		Expect(hist.GetSampleCount()).To(BeEquivalentTo(1))
		Expect(hist.GetSampleSum()).To(BeEquivalentTo(float64(len(knownBody))))
```

meyrevived (changes requested):

meyrevived (dismissed):
/lgtm

mshaposhnik (dismissed):

testing/metrics.go
```diff
@@ -0,0 +1,13 @@
+package testing
```
filariow: can we find an alternative to having this package for one function only - also just required for testing? Also this package is not under `pkg` (exported packages) nor `internal` (unexported packages)

filariow (changes requested):

testing/metrics.go
glevi-rh: I haven't thought about it really, can it introduce some issues if internal and exported packages using the same package for testing? 
Regarding to the package for one function only, please take a look at @meyrevived [comment here](https://github.com/konflux-ci/namespace-lister/pull/364#discussion_r3217636534)

filariow: I think that's un-needed complexity. We can use native gomega matchers instead of extracting the names to a list 😄 

```golang
		Expect(families).To(
			HaveField("GetName",
				ConsistOf(
					"namespace_lister_api_latency",
					"namespace_lister_api_counter",
					"namespace_lister_api_response_size",
					"namespace_lister_api_requests_in_flight",
				)))
	})
```

and 

```golang
		Expect(families).To(ContainElement(
			WithTransform(func(f *dto.MetricFamily) string {
				return f.GetName()
			}, HavePrefix("namespace_lister_process_"))))
```

meyrevived: Yup, that would also work

filariow (approved):
Mixed feeling on some unit tests, as it looks like we are testing `promhttp` to me 😄
However, let's keep these tests for now, if we see they don't work well, we'll remove or rework them

PR #371, opened 2026-05-12
KFLUXINFRA-3542: add unit tests for transform functions

glevi-rh:
## Summary
- Add unit tests for TrimAnnotations, MergeTransformFunc, TrimRole, and TrimClusterRole in internal/resourcecache/internal/transform/
- Uses gomock for MergeTransformFunc to verify call ordering without coupling to real transform logic
- Uses DescribeTable for TrimAnnotations, TrimRole, and TrimClusterRole

## Test plan
- [x] 10 tests pass locally
- [x] No changes to production code


internal/resourcecache/internal/transform/transformfuncs_test.go
```diff
@@ -0,0 +1,192 @@
...
+			// then
+			Expect(err).NotTo(HaveOccurred())
+			Expect(result).To(BeIdenticalTo(nsOut))
+		})
+
+		It("short-circuits on error", func() {
+			// given
+			tf1 := mocks.NewMockTransformFunc(ctrl)
+			tf2 := mocks.NewMockTransformFunc(ctrl)
+			tf1.EXPECT().TransformFunc(gomock.Any()).Return(nil, errors.New("boom")).Times(1)
```
filariow, line 103: we also want to enure `tf2` was not called

```suggestion
			tf1.EXPECT().TransformFunc(gomock.Any()).Return(nil, errors.New("boom")).Times(1)
			tf2.EXPECT().TransformFunc(gomock.Any()).Times(0)
```

internal/resourcecache/internal/transform/transformfuncs_test.go
```diff
@@ -0,0 +1,192 @@
...
+				}},
+			}, true, ""),
+	)
+
+	It("TrimClusterRole errors when given a non-ClusterRole object", func() {
+		_, err := transform.TrimClusterRole()(&corev1.Namespace{})
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("expected a ClusterRole"))
+	})
+})
```
filariow, line 494: we need tests for `TrimClusterRoleBinding`, `TrimRoleBinding`, and `TrimNamespace`

internal/resourcecache/internal/transform/transformfuncs_test.go
```diff
@@ -0,0 +1,192 @@
...
+
+			// then
+			Expect(err).To(MatchError("boom"))
+		})
+	})
+
+	DescribeTable("TrimRole",
+		func(role *rbacv1.Role, expectNil bool, expectErr string) {
+			result, err := transform.TrimRole()(role)
+			if expectErr != "" {
```
filariow: please split positive test cases from negative ones, so we can avoid this confusing `if`s

internal/resourcecache/internal/transform/transformfuncs_test.go
```diff
@@ -0,0 +1,192 @@
...
+			// when
+			merged := transform.MergeTransformFunc(tf1.TransformFunc, tf2.TransformFunc)
+			_, err := merged(&corev1.Namespace{})
+
+			// then
+			Expect(err).To(MatchError("boom"))
+		})
+	})
+
+	DescribeTable("TrimRole",
```
filariow: we also need to ensure ManagedFields are stripped for all the `TrimRESOURCE` functions

filariow (changes requested):

internal/resourcecache/internal/transform/transformfuncs_test.go
glevi-rh: applied

internal/resourcecache/internal/transform/transformfuncs_test.go
glevi-rh: Thanks for that, the original tests were created when they were still an anonymous functions so I missed it

filariow:
@glevi-rh Qodo has a [good point](https://github.com/konflux-ci/namespace-lister/pull/371#issuecomment-4430036401) to me (`Weak merge chaining test`)

glevi-rh:
> @glevi-rh Qodo has a [good point](https://github.com/konflux-ci/namespace-lister/pull/371#issuecomment-4430036401) to me (`Weak merge chaining test`)

I already addressed it [here](https://github.com/konflux-ci/namespace-lister/pull/371/changes#diff-11eb4bd3501213dae6b0cee09df1e330f5bf4fdfd533a7d2cce1b51a749b95f8R65-R78)

internal/resourcecache/internal/transform/transformfuncs_test.go
```diff
@@ -0,0 +1,274 @@
...
+
+	Describe("TrimRole", func() {
+		DescribeTable("valid input",
+			func(role *rbacv1.Role, expectNil bool) {
+				result, err := transform.TrimRole()(role)
+				Expect(err).NotTo(HaveOccurred())
+				if expectNil {
+					Expect(result).To(BeNil())
+					return
+				}
```
filariow: `if`s in a DescribeTable are confusing. Can we avoid them?

internal/resourcecache/internal/transform/transformfuncs_test.go
```diff
@@ -0,0 +1,274 @@
...
+				out := result.(*rbacv1.Role)
+				Expect(out.Rules).To(HaveLen(1))
+				Expect(out.Annotations).To(BeNil())
+				Expect(out.ManagedFields).To(BeEmpty())
```
filariow: we could also check that the fields we do not want to change are not changed. May you use an approach similar to what we used above in `TrimAnnotations`?

internal/resourcecache/internal/transform/transformfuncs_test.go
```diff
@@ -0,0 +1,274 @@
...
+			nsOut := nsIn.DeepCopy()
+			nsOut.Labels = map[string]string{"transformed": "true"}
+
+			tf1 := mocks.NewMockTransformFunc(ctrl)
+			tf2 := mocks.NewMockTransformFunc(ctrl)
+			gomock.InOrder(
+				tf1.EXPECT().TransformFunc(nsIn).Return(nsOut, nil).Times(1),
+				tf2.EXPECT().TransformFunc(nsOut).Return(nsOut, nil).Times(1),
+			)
+
```
filariow, line 89: please let's use something more explicit 😄 

```suggestion
			tf1 := mocks.NewMockTransformFunc(ctrl)
			tf2 := mocks.NewMockTransformFunc(ctrl)
			gomock.InOrder(
				tf1.EXPECT().TransformFunc(nsIn).Return(nsOut, nil).Times(1),
				tf2.EXPECT().TransformFunc(gomock.Any()).
					Do(func(a any) {
						if a.(*corev1.Namespace) != nsOut {
							Fail("intermediate result was not propagated as expected")
						}
					}).
					Return(nsOut, nil).
					Times(1),
			)
```

filariow (changes requested):

internal/resourcecache/internal/transform/transformfuncs_test.go
```diff
@@ -0,0 +1,416 @@
...
+			DeferCleanup(ctrl.Finish)
+		})
+
+		It("chains multiple transforms passing intermediate results", func() {
+			// given
+			nsIn := &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{Name: "test"},
+			}
+			nsOut := nsIn.DeepCopy()
+			nsOut.Labels = map[string]string{"transformed": "true"}
```
filariow: we don't need this


```suggestion
```

internal/resourcecache/internal/transform/transformfuncs_test.go
```diff
@@ -0,0 +1,416 @@
...
+						Namespace:     "default",
+						Annotations:   map[string]string{"note": "value"},
+						ManagedFields: managedFields,
+					},
+					Rules: []rbacv1.PolicyRule{{
+						APIGroups: []string{""},
+						Resources: []string{"namespaces"},
+						Verbs:     []string{"get"},
+					}},
+				}, 1),
```
filariow: we can split this in two: one for annotations one for managedFields

internal/resourcecache/internal/transform/transformfuncs_test.go
```diff
@@ -0,0 +1,416 @@
...
+	Describe("TrimRole", func() {
+		DescribeTable("strips annotations, managed fields, and non-namespace rules",
+			func(role *rbacv1.Role, expectedRuleCount int) {
+				// when
+				result, err := transform.TrimRole()(role.DeepCopy())
+
+				// then
+				Expect(err).NotTo(HaveOccurred())
+				out := result.(*rbacv1.Role)
+				Expect(out.Rules).To(HaveLen(expectedRuleCount))
```
filariow: this way we do only check quantity, not quality. If it's returning only the ones we expect it to drop, we don't fail

internal/resourcecache/internal/transform/transformfuncs_test.go
```diff
@@ -0,0 +1,416 @@
...
+			},
+			Entry("pods-only rule",
+				&rbacv1.Role{
+					ObjectMeta: metav1.ObjectMeta{Name: "pod-reader"},
+					Rules: []rbacv1.PolicyRule{{
+						APIGroups: []string{""},
+						Resources: []string{"pods"},
+						Verbs:     []string{"get"},
+					}},
+				}),
```
filariow: minor: we can reduce the length of this test by passing the Rules only and building the Role in the func body

internal/resourcecache/internal/transform/transformfuncs_test.go
```diff
@@ -0,0 +1,416 @@
...
+				result, err := transform.TrimClusterRole()(cr.DeepCopy())
+
+				// then
+				Expect(err).NotTo(HaveOccurred())
+				out := result.(*rbacv1.ClusterRole)
+				Expect(out.Rules).To(HaveLen(expectedRuleCount))
+				Expect(out.Annotations).To(BeNil())
+				Expect(out.ManagedFields).To(BeEmpty())
+
+				Expect(cr.DeepCopy()).To(
```
filariow: minor: we don't need to `DeepCopy()` here

```suggestion
				Expect(cr).To(
```

filariow (changes requested):

internal/resourcecache/internal/transform/transformfuncs_test.go
```diff
@@ -0,0 +1,434 @@
...
+				Expect(out.Annotations).To(BeNil())
+				Expect(out.ManagedFields).To(BeEmpty())
+
+				Expect(role.DeepCopy()).To(
+					WithTransform(func(r *rbacv1.Role) *rbacv1.Role {
+						r.Annotations = nil
+						r.ManagedFields = nil
+						r.Rules = out.Rules
+						return r
+					}, BeEquivalentTo(out)))
```
filariow, line 170: this kind of check can be counterintuitive, and a new reader could think it's just doing twice the same thing. Let's add some details to make it explicit.


```suggestion
				// then
				By("ensuring fields expected to be stripped were stripped")
				Expect(err).NotTo(HaveOccurred())
				out := result.(*rbacv1.Role)
				Expect(out.Rules).To(BeEquivalentTo(expectedRules))
				Expect(out.Annotations).To(BeNil())
				Expect(out.ManagedFields).To(BeEmpty())

				By("ensuring other fields were not mutated")
				Expect(role.DeepCopy()).To(
					WithTransform(func(r *rbacv1.Role) *rbacv1.Role {
						r.Annotations = nil
						r.ManagedFields = nil
						r.Rules = out.Rules
						return r
					}, BeEquivalentTo(out)))
```

This also applies for the other similar tests

internal/resourcecache/internal/transform/transformfuncs_test.go
```diff
@@ -0,0 +1,434 @@
...
+				[]rbacv1.PolicyRule{{
+					APIGroups: []string{"apps"}, Resources: []string{"namespaces"}, Verbs: []string{"get"},
+				}}),
+			Entry("empty rules", []rbacv1.PolicyRule(nil)),
+		)
+
+		It("errors when given a non-Role object", func() {
+			_, err := transform.TrimRole()(&corev1.Namespace{})
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("expected Role"))
```
filariow: we should also check the returned value is nil


```suggestion
			v, err := transform.TrimRole()(&corev1.Namespace{})
			Expect(err).To(HaveOccurred())
			Expect(err.Error()).To(ContainSubstring("expected Role"))
			Expect(v).To(BeNil())
```

filariow (changes requested):

internal/resourcecache/internal/transform/transformfuncs_test.go
glevi-rh: Addressed

internal/resourcecache/internal/transform/transformfuncs_test.go
```diff
@@ -0,0 +1,472 @@
...
+				}),
+			Entry("with managed fields",
+				&rbacv1.RoleBinding{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:          "test-rb",
+						ManagedFields: managedFields,
+					},
+				}),
+		)
+	})
```
meyrevived, line 484: So, `TrimRoleBinding` and `TrimClusterRoleBinding` are basically both aliases of the same function - they're both [defined](https://github.com/konflux-ci/namespace-lister/blob/69afb40836b8ff8dd602e17e886651df14910cc8/internal/resourcecache/internal/transform/transformfuncs.go#L90-L93) as  `trimBinding`, so we're basically testing the same thing. If `trimBinding` has a bug, both of these sets of tests are going to fail, and if `trimBinding` changes its behavior a little bit, we'll have to edit both of these sets of tests to cover the new behavior.
I'd recommend to keep one full test (let's say `TrimRoleBinding`) and reduce the other to a single `It()` that verifies `TrimClusterRoleBinding` produces the same function (or just a comment noting they're aliases).

internal/resourcecache/internal/transform/transformfuncs_test.go
```diff
@@ -0,0 +1,472 @@
...
+
+			// when
+			merged := transform.MergeTransformFunc(tf1.TransformFunc, tf2.TransformFunc)
+			result, err := merged(&corev1.Namespace{})
+
+			// then
+			Expect(err).To(MatchError("boom"))
+			Expect(result).To(BeNil())
+		})
+	})
```
meyrevived, line 147: Adding tests for some missed coverage - 

1. `MergeTransformFunc()` can be called with zero arguments. This is an edge case — the returned function should just return its input unchanged.
2. Verifying the merge degenerates correctly into a pass-through to the single function.

```suggestion
		})
		It("returns input unchanged when called with zero functions", func() {
			// given
			ns := &corev1.Namespace{
				ObjectMeta: metav1.ObjectMeta{Name: "unchanged"},
			}

			// when
			merged := transform.MergeTransformFunc()
			result, err := merged(ns)

			// then
			Expect(err).NotTo(HaveOccurred())
			Expect(result).To(BeIdenticalTo(ns))
		})

		It("delegates to a single function", func() {
			// given
			nsIn := &corev1.Namespace{
				ObjectMeta: metav1.ObjectMeta{Name: "input"},
			}
			nsOut := nsIn.DeepCopy()

			tf := mocks.NewMockTransformFunc(ctrl)
			tf.EXPECT().TransformFunc(nsIn).Return(nsOut, nil).Times(1)

			// when
			merged := transform.MergeTransformFunc(tf.TransformFunc)
			result, err := merged(nsIn)

			// then
			Expect(err).NotTo(HaveOccurred())
			Expect(result).To(BeIdenticalTo(nsOut))
		})
	})
```

internal/resourcecache/internal/transform/transformfuncs_test.go
```diff
@@ -0,0 +1,472 @@
...
+						Resources: []string{"namespaces", "pods"},
+						Verbs:     []string{"get", "list", "watch"},
+					}},
+				},
+				[]rbacv1.PolicyRule{{
+					APIGroups: []string{""},
+					Resources: []string{"namespaces"},
+					Verbs:     []string{"get"},
+				}}),
+		)
```
meyrevived, line 247: Adding tests for some missed coverage II - all TrimRole/TrimClusterRole table entries provide at most one rule that matches the namespace filter. The `filterNamespacesRelatedPolicyRules` function appends ALL matching rules to the result slice. Testing with two matching rules would verify this accumulation behavior.

```suggestion
				}}),
			Entry("preserves multiple matching rules",
				&rbacv1.Role{
					ObjectMeta: metav1.ObjectMeta{Name: "multi-match", Namespace: "default"},
					Rules: []rbacv1.PolicyRule{
						{APIGroups: []string{""}, Resources: []string{"namespaces"}, Verbs: []string{"get"}},
						{APIGroups: []string{"", "apps"}, Resources: []string{"namespaces", "pods"}, Verbs: []string{"get", "list"}},
					},
				},
				[]rbacv1.PolicyRule{
					{APIGroups: []string{""}, Resources: []string{"namespaces"}, Verbs: []string{"get"}},
					{APIGroups: []string{""}, Resources: []string{"namespaces"}, Verbs: []string{"get"}},
				}),
		)
```

internal/resourcecache/internal/transform/transformfuncs_test.go
```diff
@@ -0,0 +1,472 @@
...
+
+			// then
+			By("ensuring fields expected to be stripped were stripped")
+			Expect(err).NotTo(HaveOccurred())
+			out := result.(*corev1.Namespace)
+			Expect(out.ManagedFields).To(BeEmpty())
+			Expect(out.Spec).To(BeEquivalentTo(corev1.NamespaceSpec{}))
+			Expect(out.Status).To(BeEquivalentTo(corev1.NamespaceStatus{}))
+
+			By("ensuring other fields were not mutated")
```
meyrevived, line 430: Adding tests for some missed coverage III - `TrimNamespace()` does NOT call `TrimAnnotations()`, unlike `TrimRole()` and `TrimClusterRole()`. However, the test for `TrimNamespace` does not include annotations in the test input, so it cannot verify whether annotations are preserved or stripped.

```suggestion
					Labels:        map[string]string{"team": "infra"},
					Annotations:   map[string]string{"note": "keep-me"},
					ManagedFields: managedFields,
				},
				Spec:   corev1.NamespaceSpec{Finalizers: []corev1.FinalizerName{corev1.FinalizerKubernetes}},
				Status: corev1.NamespaceStatus{Phase: corev1.NamespaceActive},
			}

			// when
			result, err := transform.TrimNamespace()(ns.DeepCopy())

			// then
			By("ensuring fields expected to be stripped were stripped")
			Expect(err).NotTo(HaveOccurred())
			out := result.(*corev1.Namespace)
			Expect(out.ManagedFields).To(BeEmpty())
			Expect(out.Spec).To(BeEquivalentTo(corev1.NamespaceSpec{}))
			Expect(out.Status).To(BeEquivalentTo(corev1.NamespaceStatus{}))

			// TrimNamespace intentionally does NOT strip annotations, unlike TrimRole/TrimClusterRole
			By("ensuring annotations are preserved")
			Expect(out.Annotations).To(Equal(map[string]string{"note": "keep-me"}))
			
			By("ensuring other fields were not mutated")
```

meyrevived (changes requested):
Good PR, loved the use of `DescribeTable`s everywhere.
I added a few coverage bump suggestions and some comments.

internal/resourcecache/internal/transform/transformfuncs_test.go
glevi-rh: Nice catch, I'll do that.

glevi-rh: Applied in the last commit.

internal/resourcecache/internal/transform/transformfuncs_test.go
filariow: nice catch

internal/resourcecache/internal/transform/transformfuncs_test.go
```diff
@@ -0,0 +1,494 @@
...
+	})
+
+	// TrimClusterRoleBinding and TrimRoleBinding are aliases of the same
+	// trimBinding function. Full coverage is provided by TrimRoleBinding above.
+	Describe("TrimClusterRoleBinding", func() {
+		It("is the same function as TrimRoleBinding", func() {
+			Expect(reflect.ValueOf(transform.TrimClusterRoleBinding).Pointer()).To(
+				Equal(reflect.ValueOf(transform.TrimRoleBinding).Pointer()))
+		})
+	})
```
filariow, line 493: Mixed feeling on this is as it's binding to an internal detail, but I can't suggest a different approach that wouldn't duplicate the code again.
However it's smart and good for what we need. We can always change it in future if we see limitations 😄

filariow (approved):
great work team! @glevi-rh @meyrevived

meyrevived (approved):
/lgtm

PR #372, opened 2026-05-12
KFLUXINFRA-3542: add unit tests for access cache helpers

glevi-rh:
## Summary
- Move BuildAndRegisterAccessCacheMetrics and GetResyncPeriodFromEnvOrZero from root to internal/resourcecache/ and export them
- Add unit tests with DescribeTable for duration parsing (multiple valid formats + invalid input)
- Verify metrics are actually registered in the registry
- Remove export_test.go

## Test plan
- [x] 8 tests pass locally
- [x] Root package compiles cleanly with updated imports

### DIFF
--- a/internal/resourcecache/access_cache.go
+++ b/internal/resourcecache/access_cache.go
@@ -1,0 +1,44 @@
+package resourcecache
+
+import (
+	"context"
+	"os"
+	"time"
+
+	"github.com/konflux-ci/namespace-lister/internal/constants"
+	"github.com/konflux-ci/namespace-lister/internal/log"
+	"github.com/konflux-ci/namespace-lister/pkg/auth/cache"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func BuildAndRegisterAccessCacheMetrics(registry prometheus.Registerer) (cache.AccessCacheMetrics, error) {
+	if registry == nil {
+		return nil, nil
+	}
+
+	accessCacheMetrics := cache.NewAccessCacheMetrics()
+	if err := registry.Register(accessCacheMetrics); err != nil {
+		return nil, err
+	}
+	return accessCacheMetrics, nil
+}
+
+// GetValidResyncPeriodFromEnvOrZero retrieves AccessCache's ResyncPeriod from environment variables.
+// If the environment variable is not set or the value is not a valid positive duration, it returns the zero value.
+func GetValidResyncPeriodFromEnvOrZero(ctx context.Context) time.Duration {
+	var zero time.Duration
+	rps, ok := os.LookupEnv(constants.EnvCacheResyncPeriod)
+	if !ok {
+		return zero
+	}
+	rp, err := time.ParseDuration(rps)
+	if err != nil {
+		log.GetLoggerFromContext(ctx).Warn("can not parse duration from environment variable", "error", err)
+		return zero
+	}
+	if rp < 0 {
+		log.GetLoggerFromContext(ctx).Warn("negative resync period, using zero", "value", rps)
+		return zero
+	}
+	return rp
+}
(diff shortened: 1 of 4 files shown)


internal/resourcecache/access_cache_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+			// when
+			_, err = resourcecache.BuildAndRegisterAccessCacheMetrics(reg)
+
+			// then
+			Expect(err).To(BeAssignableToTypeOf(prometheus.AlreadyRegisteredError{}))
+		})
+	})
+
+	Describe("GetResyncPeriodFromEnvOrZero", func() {
```
filariow: we can't execute these tests in parallel, let's label them as `Serial`




```suggestion
	Describe("GetResyncPeriodFromEnvOrZero", Serial, func() {
```

internal/resourcecache/access_cache_test.go
```diff
@@ -0,0 +1,91 @@
...
+		const envKey = constants.EnvCacheResyncPeriod
+
+		It("returns zero when env is not set", func(ctx context.Context) {
+			// given
+			GinkgoT().Setenv(envKey, "dummy")
+			Expect(os.Unsetenv(envKey)).To(Succeed()) //nolint:usetesting // need truly-unset env for this test
```
filariow: this is good, but something shady is happening behind the hood. The `GinkgoT().Setenv` (ie. [T.Setenv](https://pkg.go.dev/testing#T.Setenv)) is deferring the `os.Setenv`. Let's make this behavior explicit and only do this at need.

```suggestion
			if v, ok := os.LookupEnv(constants.EnvCacheResyncPeriod); ok {
				// need truly-unset env for this test
				Expect(os.Unsetenv(constants.EnvCacheResyncPeriod)).To(Succeed()) //nolint:usetesting
				defer os.Setenv(constants.EnvCacheResyncPeriod, v)
			}
```

internal/resourcecache/access_cache_test.go
```diff
@@ -0,0 +1,91 @@
...
+				d := resourcecache.GetResyncPeriodFromEnvOrZero(ctx)
+
+				// then
+				Expect(d).To(Equal(expected))
+			},
+			Entry("valid duration 5m", "5m", 5*time.Minute),
+			Entry("valid duration 30s", "30s", 30*time.Second),
+			Entry("valid duration 1h", "1h", time.Hour),
+			Entry("invalid duration returns zero", "not-a-duration", time.Duration(0)),
+		)
```
filariow, line 90: We need to check for a negative duration too. This use case is not managed nowadays. We need to add some logic to handle this case, to make sure we don't break AccessCache's behavior. I think it's good to rename this method as `GetValidResyncPeriodFromEnvOrZero` and have it check that the duration is positive too. WDYT?

```suggestion
			Entry("invalid duration -1h", "-1h", ...),
		)
```

filariow (changes requested):

internal/resourcecache/access_cache_test.go
glevi-rh: Applied

filariow (dismissed):
thanks for addressing my comments. As a follow-up to this PR, we could use negative `ResyncPeriods` values to disable time-based resync at https://github.com/konflux-ci/namespace-lister/blob/849f41c0b0c4af73459a5880bfa52a26acb11eec/pkg/auth/cache/synchronized_access_cache.go#L265-L278

internal/resourcecache/access_cache.go
```diff
@@ -0,0 +1,44 @@
...
+
+	accessCacheMetrics := cache.NewAccessCacheMetrics()
+	if err := registry.Register(accessCacheMetrics); err != nil {
+		return nil, err
+	}
+	return accessCacheMetrics, nil
+}
+
+// GetValidResyncPeriodFromEnvOrZero retrieves AccessCache's ResyncPeriod from environment variables.
+// If the environment variable is not set or the value is not a valid positive duration, it returns the zero value.
```
glevi-rh:
```suggestion
// If the environment variable is not set or the value is not a valid non-negative duration, it returns the zero value.
```
@filariow that's basically what Qodo is complaining about

filariow (approved):

PR #373, opened 2026-05-12
enable e2e coverage reporting

filariow:
this enables reporting of e2e tests coverage to codecov

rh-pre-commit.version: 2.3.2
rh-pre-commit.check-secrets: ENABLED

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -121,7 +121,7 @@ jobs:
         continue-on-error: true
         run: |
           mkdir -p coverage-output && chmod 777 coverage-output
-          cp "acceptance/test/${{ matrix.proxy }}/out/namespace-lister-admin.kcfg" /tmp/kubeconfig && chmod 400 /tmp/kubeconfig
+          cp "acceptance/test/${{ matrix.proxy }}/out/namespace-lister-admin.kcfg" /tmp/kubeconfig && chmod 600 /tmp/kubeconfig
           podman run --rm \
             --network host \
             -v /tmp/kubeconfig:/kubeconfig:ro \
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -119,7 +119,7 @@ jobs:
         if: always()
         run: |
           mkdir -p coverage-output && chmod 777 coverage-output
-          cp $HOME/.kube/config /tmp/kubeconfig && chmod 644 /tmp/kubeconfig
+          cp "acceptance/test/${{ matrix.proxy }}/out/namespace-lister-admin.kcfg" /tmp/kubeconfig && chmod 644 /tmp/kubeconfig
           podman run --rm \
             --network host \
             -v /tmp/kubeconfig:/kubeconfig:ro \
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -134,11 +134,6 @@ jobs:
               --test-name=e2e-tests \
               --output=/workspace/coverage-output
 
-      - name: Warn on Collect e2e coverage failure
-        if: steps.collect_e2e_coverage.outcome == 'failure'
-        run: |
-          echo "::warning title=Coverage Collection Failed::The Collect E2E Coverage failed."
-
       - name: Upload e2e coverage to Codecov
         if: steps.collect_e2e_coverage.outcome == 'success'
         id: upload_e2e_coverage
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yaml
```diff
@@ -78,6 +78,8 @@ jobs:
   acceptance:
     name: Run acceptance tests
     runs-on: ubuntu-24.04
+    permissions:
+      id-token: write
```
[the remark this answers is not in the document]
filariow: that's not the case

.github/workflows/ci.yaml
```diff
@@ -111,6 +115,32 @@ jobs:
...
+        if: always()
+        run: |
+          mkdir -p coverage-output && chmod 777 coverage-output
+          cp "acceptance/test/${{ matrix.proxy }}/out/namespace-lister-admin.kcfg" /tmp/kubeconfig && chmod 644 /tmp/kubeconfig
+          podman run --rm \
+            --network host \
+            -v /tmp/kubeconfig:/kubeconfig:ro \
+            -v $PWD/coverage-output:/workspace/coverage-output \
+            -e KUBECONFIG=/kubeconfig \
+            quay.io/konflux-ci/konflux-devprod/coverport-cli \
```
oswcab: Can we pin this action to a version or, still better, a digest? The way it is now it;s implicitly pulling `latest` which can be non deterministic. Besides, and given we're running with `--network host` and access to a kubeconfig file, it can be a supply-chain risk.

.github/workflows/ci.yaml
```diff
@@ -111,6 +115,32 @@ jobs:
...
+            collect \
+              --namespace=namespace-lister \
+              --label-selector=apps=namespace-lister \
+              --test-name=e2e-tests \
+              --output=/workspace/coverage-output || true
```
oswcab: The issue I see when using `|| true` is that any error that's produced is not visible. I think we should rather use `continue-on-error: true` in this task so the CI is not interrupted but the errors are shown.

.github/workflows/ci.yaml
```diff
@@ -111,6 +115,32 @@ jobs:
             make -C "acceptance/test/${{ matrix.proxy }}" test
           echo "##[endgroup]"
 
+      - name: Collect e2e coverage
+        if: always()
+        run: |
+          mkdir -p coverage-output && chmod 777 coverage-output
```
oswcab, line 123: Do we really need 777? Looks like very open and permissive... I know these are ephemeral runners, but... could 775 (non world-writable) enough in this case?

.github/workflows/ci.yaml
```diff
@@ -111,6 +115,32 @@ jobs:
...
+          cp "acceptance/test/${{ matrix.proxy }}/out/namespace-lister-admin.kcfg" /tmp/kubeconfig && chmod 644 /tmp/kubeconfig
```
oswcab: Given the file is mounted as `read-only` below, shouldn't be enough `chmod 600` here?

oswcab (commented):
A few nits, otherwise, LGTM

.github/workflows/ci.yaml
filariow: I had the [same concern here](https://github.com/konflux-ci/may/pull/114#discussion_r3181180621)

.github/workflows/ci.yaml
filariow: good catch, [400 is good too](https://github.com/konflux-ci/may/pull/114#discussion_r3181186104)

filariow:
```suggestion
          cp "acceptance/test/${{ matrix.proxy }}/out/namespace-lister-admin.kcfg" /tmp/kubeconfig && chmod 400 /tmp/kubeconfig
```

psturc:
the coverage data collection currently [fails on loading kubeconfig](https://github.com/konflux-ci/namespace-lister/actions/runs/25787587731/job/75744659395?pr=373#step:6:25)

.github/workflows/ci.yaml
filariow: I don't like the fact that the [error is suppressed neither](https://github.com/konflux-ci/namespace-lister/pull/373#issuecomment-4439831691). At the same time, I don't want the failure of the coverage upload step to prevent us from merging a PR. `continue-on-error` should fulfill both requirements, let's use it 😄

psturc (approved):

.github/workflows/ci.yaml
filariow: It looks like it's not. It won't work unless you set 644. Look at the commit history 🙃

meyrevived (approved):
/lgtm

PR #383, opened 2026-05-20
feat: add health and metrics acceptance test suites

oswcab:
Add BDD tests for service health and observability:
- Liveness endpoint returns 200
- Readiness endpoint returns 200
- Metrics service is available in the cluster

Contributes to: KFLUXINFRA-3243

### DIFF
--- a/acceptance/pkg/context/context.go
+++ b/acceptance/pkg/context/context.go
@@ -2,6 +2,7 @@ package context
 
 import (
 	"context"
+	"crypto/tls"
 
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -10,10 +11,12 @@ import (
 type ContextKey string
 
 const (
-	ContextKeyNamespaces      ContextKey = "namespaces"
-	ContextKeyRunId           ContextKey = "run-id"
-	ContextKeyUserInfo        ContextKey = "user-info"
-	ContextKeyBuildUserClient ContextKey = "build-user-client"
+	ContextKeyNamespaces             ContextKey = "namespaces"
+	ContextKeyRunId                  ContextKey = "run-id"
+	ContextKeyUserInfo               ContextKey = "user-info"
+	ContextKeyBuildUserClient        ContextKey = "build-user-client"
+	ContextKeyNamespaceListerAddress ContextKey = "namespace-lister-address"
+	ContextKeyTLSConfig ContextKey = "tls-config"
 )
 
 type BuildUserClientFunc func(context.Context) (client.Client, error)
@@ -48,6 +51,22 @@ func WithRunId(ctx context.Context, runId string) context.Context {
 
 func RunId(ctx context.Context) string {
 	return get[string](ctx, ContextKeyRunId)
+}
+
+func WithNamespaceListerAddress(ctx context.Context, address string) context.Context {
+	return into(ctx, ContextKeyNamespaceListerAddress, address)
+}
+
+func NamespaceListerAddress(ctx context.Context) string {
+	return get[string](ctx, ContextKeyNamespaceListerAddress)
+}
+
+func WithTLSConfig(ctx context.Context, cfg *tls.Config) context.Context {
+	return into(ctx, ContextKeyTLSConfig, cfg)
+}
+
+func TLSConfig(ctx context.Context) *tls.Config {
+	return get[*tls.Config](ctx, ContextKeyTLSConfig)
 }
 
 // aux
--- a/acceptance/pkg/suite/steps_health.go
+++ b/acceptance/pkg/suite/steps_health.go
@@ -1,0 +1,61 @@
+package suite
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"time"
+
+	"github.com/cucumber/godog"
+
+	tcontext "github.com/konflux-ci/namespace-lister/acceptance/pkg/context"
+)
+
+func InjectHealthSteps(ctx *godog.ScenarioContext) {
+	ctx.Then(`^the healthz endpoint returns 200$`, healthzReturns200)
+	ctx.Then(`^the readyz endpoint returns 200$`, readyzReturns200)
+}
+
+func healthzReturns200(ctx context.Context) (context.Context, error) {
+	return checkEndpoint(ctx, "/healthz")
+}
+
+func readyzReturns200(ctx context.Context) (context.Context, error) {
+	return checkEndpoint(ctx, "/readyz")
+}
+
+func checkEndpoint(ctx context.Context, path string) (context.Context, error) {
+	address := tcontext.NamespaceListerAddress(ctx)
+	if address == "" {
+		return ctx, fmt.Errorf("namespace-lister address not set in context")
+	}
+
+	tlsConfig := tcontext.TLSConfig(ctx)
+	client := &http.Client{
+		Timeout: 10 * time.Second,
+		Transport: &http.Transport{
+			TLSClientConfig: tlsConfig,
+		},
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, address+path, nil)
+	if err != nil {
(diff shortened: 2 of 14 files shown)


acceptance/pkg/suite/steps_health.go
```diff
@@ -0,0 +1,61 @@
...
+	tlsConfig := tcontext.TLSConfig(ctx)
+	client := &http.Client{
+		Timeout: 10 * time.Second,
+		Transport: &http.Transport{
+			TLSClientConfig: tlsConfig,
+		},
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, address+path, nil)
+	if err != nil {
```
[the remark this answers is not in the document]
meyrevived: That is a good point @oswcab

acceptance/test/smart-proxy/hook_test.go
```diff
@@ -17,12 +18,21 @@ func InjectHooks(ctx *godog.ScenarioContext) {
...
 
 func injectBuildUserClient(ctx context.Context, sc *godog.Scenario) (context.Context, error) {
 	return tcontext.WithBuildUserClientFunc(ctx, buildUserClientForAuthProxy), nil
 }
 
+func injectServiceAddresses(ctx context.Context, sc *godog.Scenario) (context.Context, error) {
+	address := suite.EnvKonfluxAddressOrDefault(defaultTestAddress)
+	ctx = tcontext.WithNamespaceListerAddress(ctx, address)
+	ctx = tcontext.WithMetricsAddress(ctx, address)
+	ctx = tcontext.WithTLSConfig(ctx, &tls.Config{InsecureSkipVerify: true}) //nolint:gosec
```
meyrevived: Why hardcode `tls.Config{InsecureSkipVerify: true}` and not do it like in [config.go](https://github.com/konflux-ci/namespace-lister/blob/63f733934e7d065017c6068e89df903857f2b21d/acceptance/pkg/rest/config.go#L9-L17)?
Any regression in TLS configuration can never be caught by the tests this way.

acceptance/test/dumb-proxy/hook_test.go
```diff
@@ -17,14 +18,22 @@ func InjectHooks(ctx *godog.ScenarioContext) {
...
 
 func injectBuildUserClient(ctx context.Context, sc *godog.Scenario) (context.Context, error) {
 	return tcontext.WithBuildUserClientFunc(ctx, buildUserClientWithTokenReview), nil
 }
 
+func injectServiceAddresses(ctx context.Context, sc *godog.Scenario) (context.Context, error) {
+	address := suite.EnvKonfluxAddressOrDefault(defaultTestAddress)
+	ctx = tcontext.WithNamespaceListerAddress(ctx, address)
+	ctx = tcontext.WithMetricsAddress(ctx, address)
+	ctx = tcontext.WithTLSConfig(ctx, &tls.Config{InsecureSkipVerify: true}) //nolint:gosec
```
meyrevived: Same comment as for [this](https://github.com/konflux-ci/namespace-lister/pull/383/changes#r3280350634) instance of config hardcoding

acceptance/pkg/context/context.go
```diff
@@ -50,6 +55,46 @@ func RunId(ctx context.Context) string {
...
+}
+
+func GetHTTPResponse(ctx context.Context) HTTPResponse {
+	return get[HTTPResponse](ctx, ContextKeyHTTPResponse)
+}
+
+type HTTPResponse struct {
+	StatusCode int
+	Body       []byte
+}
```
meyrevived: `ContextKeyHTTPResponse`, `WithHTTPResponse()`, `GetHTTPResponse()`, and the `HTTPResponse` struct are defined but never used anywhere in the code within this PR - is there a file missing or something?

acceptance/pkg/suite/steps_metrics.go
```diff
@@ -0,0 +1,45 @@
...
+
+	arest "github.com/konflux-ci/namespace-lister/acceptance/pkg/rest"
+)
+
+func InjectMetricsSteps(ctx *godog.ScenarioContext) {
+	ctx.Then(`^the metrics service is available in the cluster$`, metricsServiceIsAvailable)
+}
+
+func metricsServiceIsAvailable(ctx context.Context) (context.Context, error) {
+	cli, err := arest.BuildDefaultHostClient()
```
meyrevived, line 20: The metrics step uses `arest.BuildDefaultHostClient()` to query the cluster directly rather than using the stored metrics address. The context value from  `acceptance/pkg/context/context.go:78-80` is set but orphaned.

Either use `MetricsAddress(ctx)` in `metricsServiceIsAvailable` to hit the actual metrics endpoint, or remove the setter/getter pair if the step intentionally checks the Kubernetes Service object instead.

acceptance/test/dumb-proxy/hook_test.go
```diff
@@ -17,14 +18,22 @@ func InjectHooks(ctx *godog.ScenarioContext) {
 	suite.InjectBaseHooks(ctx)
 
 	ctx.Before(injectBuildUserClient)
+	ctx.Before(injectServiceAddresses)
 }
 
 func injectBuildUserClient(ctx context.Context, sc *godog.Scenario) (context.Context, error) {
```
meyrevived, line 23: This whole code block is identicle between the dumb-proxy and the smart-proxy suites, as are the health.features and metrics.feature.
Duplicate code is a problem because if something in the shared code breaks, both test suites fail and we get x2 the failure rate. Also, if we change something in the shared code, we have to make adjustments to the tests in two places.
If if godog's test runner doesn't require features to be co-located with the test package, let's DRY it up.

meyrevived (changes requested):
Thank you for making this effort to set up good and robust testing infrastructure!
I suggested some refactoring for better maintainability and found a couple of issues.

acceptance/pkg/context/context.go
oswcab: Indeed. Done.

acceptance/pkg/suite/steps_health.go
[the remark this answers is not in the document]
oswcab: Done.

acceptance/pkg/suite/steps_metrics.go
oswcab: Done.

acceptance/test/dumb-proxy/hook_test.go
oswcab: Extracted `InjectServiceAddresses()` into the shared suite package, eliminating the duplicated `injectServiceAddresses` function from both hook files. The feature files (`health.feature`, `metrics.feature`) remain duplicated because Go's `//go:embed` requires files to be in the package directory tree.

acceptance/test/dumb-proxy/hook_test.go
oswcab: Done.

acceptance/test/smart-proxy/hook_test.go
oswcab: Done.

meyrevived (approved):
/lgtm

acceptance/test/dumb-proxy/config/proxy/nginx.conf
```diff
@@ -57,6 +57,16 @@ http {
...
+        location = /healthz {
+            proxy_pass https://namespace-lister.namespace-lister.svc.cluster.local:12000;
+            proxy_read_timeout 1m;
+        }
+
+        location = /readyz {
+            proxy_pass https://namespace-lister.namespace-lister.svc.cluster.local:12000;
+            proxy_read_timeout 1m;
+        }
+
```
filariow, line 69: these endpoints could be served through the proxy, but it's not technically the best idea. They are meant for internal consumption. To simplify tests we can reflect this "suggestion" and do not expose them.

Also a quick way to test these without introducing complex e2e tests is to simply add `liveness` and `readyness` probes to the Deployment. The namespace-lister won't come up if they are not working.

PR #386, opened 2026-05-22
fix: smart-proxy unauthenticated test was silently passing

oswcab:
The `userRequestIsRejectedWithUnauthorizerError` assertion had a bug: when the request succeeded (err == nil), the check `!errors.IsUnauthorized(nil)` evaluated to true and executed `return ctx, err` — which returned nil, silently passing the test.

This meant the smart-proxy "user not authenticated" scenario (list.feature:13) never actually verified a 401 response. The unauthenticated client built by `buildUnauthenticatedUserClientForAuthProxy` retained the admin kubeconfig credentials (bearer token and client certs), so requests were authenticated via TokenReview as the admin user — returning an empty namespace list instead of 401.

Two options were considered:
1. Change the scenario expectation to accept an empty list (since the client was technically authenticated as admin with no bindings).
2. Fix the unauthenticated client to truly strip credentials so TokenReview rejects the request with 401.

Option 2 was chosen because the scenario's intent is to test unauthenticated access. Stripping bearer token, token file, client cert, and client key from the config ensures the request carries no credentials, making TokenReview reject it as expected.

Contributes to: KFLUXINFRA-3342

### DIFF
--- a/acceptance/test/smart-proxy/hook_test.go
+++ b/acceptance/test/smart-proxy/hook_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/cucumber/godog"
+	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	tcontext "github.com/konflux-ci/namespace-lister/acceptance/pkg/context"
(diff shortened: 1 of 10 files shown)


acceptance/test/smart-proxy/hook_test.go
```diff
@@ -39,14 +39,19 @@ func buildUserClientForAuthProxy(ctx context.Context) (client.Client, error) {
...
-	// the client should be unauthenticated,
-	// so skip any impersonation here
+	// strip all credentials so the request is truly unauthenticated;
+	// without a valid token or client cert, TokenReview will reject the request
+	cfg.BearerToken = ""
+	cfg.BearerTokenFile = ""
+	cfg.CertData = nil
+	cfg.CertFile = ""
+	cfg.KeyData = nil
+	cfg.KeyFile = ""
```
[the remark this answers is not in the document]
oswcab: Done

acceptance/test/smart-proxy/hook_test.go
```diff
@@ -44,6 +44,18 @@ func buildUnauthenticatedUserClientForAuthProxy(ctx context.Context) (client.Cli
 		return nil, err
 	}
 
+	// strip all credentials so the request is truly unauthenticated
```
filariow: what if instead of deny-listing, we allow listing? we create a new empty cfg and copy only the allowed ones from `NewDefaultClientConfig`'s one?

filariow (changes requested):

acceptance/test/smart-proxy/hook_test.go
oswcab: Good call. Done.

